PR #6, opened 2021-07-14
Added collective pallet and made validator set work with collective origin.

gautamdhameja:


stiiifff:
Gave it a try to pass a motion to add Charlie as validator to Validator set, and I got a `BadOrigin` error when motion was executed.

Also, I could vote with Validator that was being added to validator set ... maybe we would want to exclude that account from the vote ?

Rundown:
- Alice proposes motion (and votes Aye)
- Bob & Charlie vote Aye
- Dave & Ferdie vote Nay
- Proposal reaches threshold
- Motion is approved
- Motion execution fails

image
image
image

gautamdhameja:
Oh, let me check. Maybe the origin setup is incorrect.

gautamdhameja:
I have made a couple of changes and then it worked for me.

Before running - Alice and Bob are set as initial authorities and also as council members.

1. Start the chain in the local_testnet mode with the non-validator (`charlie`) first (so it take the default port and connects with the web app).
2. Start `alice` and `bob` nodes too.
3. Alice and Bob will produce blocks and Charlie will only import them as it is not a validator yet.
4. From Charlie's node connected web app instance, make an RPC call `author - rotate_keys`. This will print session keys for `Charlie` node.
5. Use this printed string in the step 4 and make an extrinsic call from Charlie account - `Session - set_keys`. Put `0x` in the proof input.
6. Now start the council flow. From Alice propose a motion to add Charlie as a validator using the validator-set pallet. (I used 2 as threshold and 200 as length.)
7. From Bob account, vote and approve this proposal.
8. Wait for the voting period to end. Close the motion using Alice account.
9. The motion will execute and Charlie will be added as validator. (see screenshot below)

Screenshot 2021-07-15 at 15 43 19

stiiifff:
I'm an idiot .. completely forgot points 1 to 5 before proposing the motion ... while I did go through them to test last time.
Maybe it was working and you didn't need to make any change ? Let me re-check that.

gautamdhameja:
I had to make a small change to the council genesis config but that was not related to the error you got.

stiiifff:
I'm testing .. I think I know why.

stiiifff:
Ok, it has to do with the `EnsureProportionMoreThan` and the threshold specified for the motion. If the threshold specified is less or equal to the half of the council has required by the origin, the motion execution fails though it was approved as per the treshold.

stiiifff:
Maybe we can use `EnsureProportionAtLeast` instead to make it foolproof in the context of this sample.

gautamdhameja:
Yeah, makes sense. Done.

stiiifff (approved):

Issue #56, opened 2022-09-05
Execute a XCM call from an ink! contract.


hbulgarini:
Implement the [pallet-contracts-xcm](https://github.com/paritytech/pallet-contracts-xcm) chain extension and execute a simple XCM message from a smart contract developed in ink!

PR #60, opened 2022-09-16
Executing a XCM call from an ink! contract.

hbulgarini:
Closes issue #56 

This PR is a PoC of the [xcm chain extension](https://github.com/paritytech/pallet-contracts-xcm) mvp that allow us to send (and receive) XCM messages from an ink! contract.

Also i have included a new folder which is the `contracts` folder where i believe this can be used as a common place to drop all the future contracts that we might develop.

The contract is very simple it self, it just build a simple Transact that can be sent to any parachain under the same parent (in XCM terms). 

P.S: i have many issues with the versioning around the chain extension as where as the contract, so i think there is going to be a more stable version once the contracts team puts this MVP into a more productive release.


stiiifff (approved):

stiiifff:
Nice ! Looking forward to see it in action.

stiiifff (approved):

PR #78, opened 2022-11-09
44 defi xcm use case no1

lana-shanghai:
Generate two different runtimes from the same repo with the features flag.


stiiifff:
@lana-shanghai can you investigate the CI build issue ? thx !

lana-shanghai:
> @lana-shanghai can you investigate the CI build issue ? thx !

This is because the commands to build have to include the flag. The commands are in README. Quote: 
`cargo b -r --features with-base-runtime`
`cargo b -r --features with-trappist-runtime`

hbulgarini:
> > @lana-shanghai can you investigate the CI build issue ? thx !
> 
> This is because the commands to build have to include the flag. The commands are in README. Quote: `cargo b -r --features with-base-runtime` `cargo b -r --features with-trappist-runtime`

quick question: what about if i want to build both? 

should i execute `cargo b -r --features with-trappist-runtime with-base-runtime` ?

bin/polkadot
```diff
@@ -0,0 +1 @@
+../../../everything-trappist/trappist/bin/polkadot
```
hbulgarini, line 1: We should remove bin folder and just let the developer choosing their own bin locations.

build_runtimes.sh
```diff
@@ -0,0 +1,3 @@
+cargo b -r --features with-base-runtime
+cp target/release/trappist-collator bin/base-collator
```
hbulgarini, line 2:
```suggestion
mc target/release/trappist-collator target/release/base-collator
```

instead of copying to bin, i would just rename it in the same target/release folder so we have both collators under the same folder (kind of expected)

node/service/src/chain_spec/base.rs
```diff
@@ -0,0 +1,247 @@
...
+			key: Some(root_key),
+		},
+		assets: AssetsConfig {
+			assets: vec![
+				// id, owner, is_sufficient, min_balance
+				(LOCAL_ASSET_ID, get_account_id_from_seed::<sr25519::Public>("Alice"), true, 1),
+			],
+			metadata: vec![
+				// id, name, symbol, decimals
+				(LOCAL_ASSET_ID, "bUSD".into(), "BSD".into(), 10),
```
hbulgarini, line 235:
```suggestion
				(LOCAL_ASSET_ID, "xUSD".into(), "BASE".into(), 10),
```

There is a stablecoin already called bUSD :/

hbulgarini (commented):
I left some comments.

zombienet-macos
```diff
@@ -0,0 +1 @@
+../../everything-trappist/trappist/zombienet-macos
```
hbulgarini, line 1: can we remove this symbolic link?

hbulgarini (changes requested):

bin/polkadot
lana-shanghai: Will do

build_runtimes.sh
lana-shanghai: Agree

hbulgarini (dismissed):
LGTM!

hbulgarini (approved):
LGTM!

PR #79, opened 2022-11-10
Custom Matcher (ConvertedRegisteredAssetId) for ReservedFungiblesTransactor

### LINKED ISSUE #51: Asset registry
Currently, Trappist's XCM configuration makes a bold assumption that the local's Assets pallet will be configured with the same asset ID as assets on Statemine which acts as reserve for any assets configured over there: https://github.com/paritytech/trappist/blob/2ef44e00e1abed0c02aeeffb4e4bdb5210141908/runtime/src/xcm_config.rs#L119-L143. This is a rather simplistic (and error-prone) logic.

- Moonbeam uses an AssetManager pallet: https://github.com/PureStake/moonbeam/blob/master/pallets/asset-manager/src/lib.rs
- Acala uses an AssetRegistry pallet (from ORML): https://github.com/open-web3-stack/open-runtime-module-library/tree/master/asset-registry; an the Unknown Tokens pallet (ORML): https://github.com/open-web3-stack/open-runtime-module-library/tree/master/unknown-tokens

We should implement a mechanism similar to an asset manager / registry so that only registered foreign assets are considered for reserve-based transfer operations.

bernardoaraujor:
close #51 

`ConvertedRegisteredAssetId` is a custom implementation of the `MatchesFungibles` trait that allows `ReservedFungiblesTransactor` to coexist with `LocalFungiblesTransactor` without having its execution priority block the second Transactor from ever being triggered.

Whenever `ConvertedRegisteredAssetId` can't find the Asset in `AssetRegistry` (`what = ConvertAssetId::convert_ref(id)`), a `MatchError::AssetNotFound` is thrown (instead of `MatchError::AssetIdConversionFailed`).

That allows the [`TransactAsset` trait implementation for Tuple](https://github.com/paritytech/polkadot/blob/3711c6f9b2a30ab6cc888917ad0bbfa64266d41b/xcm/xcm-executor/src/traits/transact_asset.rs#L117) to simply bypass the `ReservedFungiblesTransactor` without halting the XCM execution with a `FailedToTransactAsset("AssetIdConversionFailed")` error.

Integration tests that assert the correctness of this solution can be found on the [`bar/52-asset-trap`](https://github.com/paritytech/trappist/tree/bar/52-asset-trap/integration-tests/asset-trap) branch.

### DIFF
--- a/integration-tests/asset-registry/0_reserve_transfer.yml
+++ b/integration-tests/asset-registry/0_reserve_transfer.yml
@@ -227,7 +227,7 @@ tests:
                           - type: XcmV2TraitsOutcome
                             xcmOutcome: Complete
                             threshold: [10, 10]
-                            value: 654,608,000
+                            value: 1,000,000,000
                       - name: assets.Transferred
                         attributes:
                           - type: AccountId32
--- a/primitives/xcm/src/lib.rs
+++ b/primitives/xcm/src/lib.rs
@@ -1,7 +1,8 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use sp_std::{borrow::Borrow, marker::PhantomData};
-use xcm::latest::MultiLocation;
+use xcm::latest::{AssetId::Concrete, Fungibility::Fungible, MultiAsset, MultiLocation};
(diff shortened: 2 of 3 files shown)


stiiifff (approved):
LGTM

PR #82, opened 2022-11-11
Cache build artefacts in CI build

### LINKED ISSUE #83: CI build is slow
The CI build is a bit slow as the benchmarking phase is not re-using build artefacts of the build step.
We should use the [cache](https://github.com/actions/cache) Github action to solve that.

stiiifff:
Fixes #83


evilrobot-01:
Great to see! 

Just wanted to mention that we _may_ need to fix the version of nightly to prevent cache potentially becoming ineffective each day, if it matters.

An example from a previous file I implemented elsewhere:

```
      # Install and use nightly
      - name: Install nightly toolchain
        uses: actions-rs/toolchain@v1
        with:
          profile: minimal
          toolchain: nightly-2022-09-01 # Specific version required otherwise new cache each day due to nightly changes
          override: true
          components: rustfmt, clippy

      # Setup cache
      - name: Setup cache
        uses: Swatinem/rust-cache@v2
        with:
          shared-key: "checks-v2"
```

evilrobot-01:
Might need the same toolchain in the formatting check step?

```
      - name: Check formatting
        uses: actions-rs/cargo@v1
        with:
          toolchain: nightly-2022-11-02
          command: fmt
          args: --all --check
```

evilrobot-01:
Failed, think the caching needs to be made aware of the wasm artifacts. Will take a look after call.

evilrobot-01:
Seems a subsequent run (using cache) is looking for wasm runtimes, even though they aren't generated due to SKIP_WASM_BUILD=1. Attempting without the flag as a quick fix.

evilrobot-01 (approved):
Had hoped to get it down to around 10 mins, but half the standard build time is an improvement nonetheless.

PS - there is apparently another Parity CI tool called `rusty-cachier` which may further improve this, so hopefully we can adopt in the future.

PR #87, opened 2022-11-14
Edit target dir for base runtime

lana-shanghai:
Separated the target folders for different runtimes due to conflicting build artefacts.


node/service/src/chain_spec/base.rs
```diff
@@ -230,7 +230,7 @@ fn testnet_genesis(
 			],
 			metadata: vec![
 				// id, name, symbol, decimals
-				(LOCAL_ASSET_ID, "BASE".into(), "xUSD".into(), 10),
```
bernardoaraujor, line 233: I think this can be dropped.
See https://github.com/paritytech/trappist/pull/81

lana-shanghai: OK will remove it here as well, thanks Bernardo!

stiiifff:
(oops, was a bit too quick to merge)

PR #91, opened 2022-11-15
Suggestion: relocate build script

evilrobot-01:
Minor suggestion, relocate runtime build script to existing scripts directory.

`./scripts/build_runtime.sh`


lana-shanghai (approved):
Yes, makes sense! LGTM

PR #92, opened 2022-11-16
Remove ReservedFungiblesTransactor (base runtime)

evilrobot-01:
- removed `ReservedFungiblesTransactor` from `AssetTransactors`
- copied updated definitions of `CurrencyTransactor` and `FungiblesTransactor` from [Penpal](https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/testing/penpal/src/xcm_config.rs) implementation, including new `NonZeroIssuance` for teleports
- removed disused pallets:
  - pallet-asset-registry: removed from base-runtime
  - pallet-society: seems to have snuck into cargo.toml at some point but wasnt being used in code

Double hop reserve-transfer tested successfully using resulting base-runtime (https://github.com/paritytech/parachains-integration-tests/tree/frank/trappist/examples/trappist/reserve-transfer-asset)


runtime/base/src/xcm_config.rs
```diff
@@ -81,67 +76,44 @@ pub type LocationToAccountId = (
...
 	// Convert an XCM MultiLocation into a local account id:
 	LocationToAccountId,
 	// Our chain's account ID type (we can't get away without mentioning it explicitly):
 	AccountId,
-	// We don't track any teleports of `Assets`.
-	Nothing,
-	// We don't track any teleports of `Assets`.
+	// We only want to allow teleports of known assets. We use non-zero issuance as an indication
+	// that this asset is known.
+	NonZeroIssuance<AccountId, Assets>,
```
bernardoaraujor, line 110: do we have an integration test on teleporting assets?
it seems we've only tested teleports on native tokens so far... tbh I don't really have a mental model on how teleports would work on assets

evilrobot-01: Not yet - took a quick look at manually teleporting xUSD from Statemine to Trappist/Base(3000) and both received the xcm message, but failed with:

```
xcmpQueue.Fail
error: XcmV2TraitsError
UntrustedTeleportLocation
```

I then tweaked the config of the `base-runtime` to match that of the reserve-transfer setup and seemed to be able to successfully teleport xUSD to the 'Base Local' chain.

```
impl xcm_executor::Config for XcmConfig {
	type IsReserve = Reserves;
-	type IsTeleporter = (); // Teleporting is disabled.
+	type IsTeleporter = Reserves;

impl pallet_xcm::Config for Runtime {
-	type XcmTeleportFilter = Nothing;
+	type XcmTeleportFilter = Everything;
	type XcmReserveTransferFilter = Everything;
```

image

Shall I update this PR with these changes, and then add a quick integration test to the other PR?

stiiifff: Teleporting only makes sense between two different consensus systems (e.g. ETH <-> Polkadot) or between the relay chain and a common-good parachain (e.g. Polkadot -> Statemint), and there must exists a trust relationship (the sending chain trusts the receiving one to do proper accounting and not to mess up the token's emission / supply).

stiiifff: So in the context of Trappist, the only Teleport use case we have is for teleporting the relay chain's (Rococo) native token to Statemine.

evilrobot-01: Thanks for clarifying!

evilrobot-01: I'll remove it 👍

evilrobot-01: Should be good to go. 

Also fixed the build script (in line with GH workflow changes) after the recent PR to set trappist-runtime as default feature. It was a quick fix so didnt think it warranted a separate PR, but can revert if you'd prefer.

bernardoaraujor (approved):

PR #93, opened 2022-11-17
Build trappist runtime by default

stiiifff:
Build failing when running `cargo build --release` is counter-intuitive.
Build the trappist-runtime by default (node's default feature).


evilrobot-01 (approved):
LGTM

PR #108, opened 2023-01-12
fix(base): add local asset transactor type definition

evilrobot-01:
Additional type aliases defined within mock base runtime to correctly resolve types, rather than re-using from actual runtimes.

Also added small helper function for optionally outputting events, useful when debugging.

### DIFF
--- a/xcm-simulator/src/lib.rs
+++ b/xcm-simulator/src/lib.rs
@@ -210,13 +210,12 @@ mod tests {
 	use crate::relay_chain::mock_paras_sudo_wrapper;
 	use codec::Encode;
 	use frame_support::{
-		assert_ok,
+		assert_ok, log,
 		pallet_prelude::{DispatchResult, DispatchResultWithPostInfo},
 		traits::PalletInfoAccess,
 	};
 	use std::sync::Once;
 	use thousands::Separable;
-	use tracing::trace;
 	use xcm::prelude::*;
 	use xcm_simulator::{TestExt, Weight};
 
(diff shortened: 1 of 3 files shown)


xcm-simulator/src/lib.rs
```diff
@@ -589,6 +589,36 @@ mod tests {
...
+	fn event_collection_works() {
+		init_tracing();
+
+		MockNet::reset();
+
+		const AMOUNT: u128 = trappist::EXISTENTIAL_DEPOSIT * 10;
+		const MAX_WEIGHT: u128 = 1_000_000_000;
+
+		Trappist::execute_with(|| {
+			assert_ok!(trappist::PolkadotXcm::execute(
```
bernardoaraujor, line 602: If the goal is simply to emit an event, it's not clear why we're actually doing a `WithdrawAsset` XCM here.

Something like this makes the intention more explicit:
```
assert_ok!(trappist::System::remark_with_event(trappist::RuntimeOrigin::signed(ALICE), vec![1]));
```

xcm-simulator/src/lib.rs
```diff
@@ -589,6 +589,36 @@ mod tests {
...
+				Weight::from_ref_time(MAX_WEIGHT as u64)
+			));
+			output_events::<trappist::Runtime>();
+			assert_eq!(3, trappist::System::events().len());
+		});
+
+		Base::execute_with(|| {
+			assert_ok!(base::PolkadotXcm::execute(
+				base::RuntimeOrigin::signed(ALICE),
+				Box::new(VersionedXcm::from(Xcm(vec![WithdrawAsset(((0, Here), AMOUNT).into())]))),
```
bernardoaraujor, line 614: same thing

xcm-simulator/src/lib.rs
evilrobot-01: The only reason for these is that they actually trigger multiple events, most importantly from the balances pallet when fees are charged, which was the issue you encountered and this PR addresses. 

Perhaps a simple rename of the test to something like `execute_deposits_events`?

bernardoaraujor: ok, I was missing that point, that makes sense then.

bernardoaraujor (approved):

PR #109, opened 2023-01-12
TrappistDropAssets XCM primitive

### LINKED ISSUE #52: XCM Config: Asset trap & claim system
Currently, Trappist's XCM configuration relies on the default behavior for asset-related errors. We might want to explain & show how the asset trap & claim systems work in XCM learning sessions / workshops.

As an example, Acala implements custom logic for the asset trap: https://github.com/AcalaNetwork/Acala/blob/47e6600483976b7992b5c391a3949e6d66b7b937/runtime/acala/src/xcm_config.rs#L173-L180; https://github.com/AcalaNetwork/Acala/blob/master/runtime/common/src/xcm_impl.rs#L84-L126

See Gav's original post on XCM error management: https://polkadot.network/blog/xcm-part-three-execution-and-error-management/

bernardoaraujor:
close #52 

The assets to be trapped are checked. Non-compliant assets will not be trapped (they're lost forever). The asset will be trapped if (and only if):
- It is Trappist's Native Token (`HOP`) OR some derivative asset previously registered in `pallet-asset-registry` (e.g.: `txUSD`).
- The amount is bigger than the token's minimum balance (a.k.a. existential deposit).

`pallet-xcm` is still used as the ultimate asset trapping mechanism (stores the hashes).

Asset claiming remains unchanged.

Remains as ToDo: benchmark execution of `drop_assets` and initialize `weight` variable with real weight.


bernardoaraujor:
@evilrobot-01 I split the xcm-simulator tests into different files.

The goal to avoid `lib.rs` from growing too big, as well as mixing tests of different contexts into the same file without a clear separation between them.

I'm not sure my solution was ideal however, because I ended up having to re-use a lot of code in both files.

That could possibly cause conflicts with your recent PRs https://github.com/paritytech/trappist/pull/107 + https://github.com/paritytech/trappist/pull/108

Also not sure how CI will behave. Let me know if you have any strong opinions about this. Happy to pivot the strategy as requested.

bernardoaraujor:
Converting back into draft as currently, while each test module works as executed individually, they're breaking when run all together via `cargo test tests::`

primitives/xcm/src/lib.rs
```diff
@@ -49,3 +58,66 @@ impl<
...
+					}
+				}
+			}
+		}
+
+		// TODO: put real weight of execution up until this point here
+		let mut weight = 0;
+
+		if !trap.is_empty() {
+			weight += XcmPallet::drop_assets(origin, trap.into());
```
evilrobot-01, line 118: Maybe add a simple comment here so easier for newcomers (like me) to understand the reason for this: "Deposit remaining assets in holiding registry into asset trap, for later claiming via ClaimAsset instruction"

primitives/xcm/src/lib.rs
```diff
@@ -49,3 +58,66 @@ impl<
 		Ok((what, amount))
 	}
 }
+
+pub struct TrappistDropAssets<
```
evilrobot-01, line 62: TrappistAssetTrap might be a slightly better name, mostly as I associate drop with a database drop statement so find it a little confusing (i.e. destroying assets)

evilrobot-01 (commented):
A few minor suggestions for consistency and clarity, let me know what you think.

primitives/xcm/src/lib.rs
bernardoaraujor: The rationale behind this name is to signal that this is Trappist's implementation of the `DropAssets` trait.

I think the term drop comes from "drop Holding Register".

Conceptually speaking, the Asset Trap is composed of both `DropAssets` and `ClaimAssets` traits. So `TrappistAssetTrap` wouldn't encapsulate that meaning.

Our `XcmConfig` is using `TrappistDropAssets` as the implementation for `DropAssets`, but `pallet-xcm` for `ClaimAssets`.

evilrobot-01: Makes sense now, thanks!

integration-tests/asset-trap/1_trap_claim_native_asset.yml
```diff
@@ -0,0 +1,165 @@
...
+                              {"AccountId32":
+                                {
+                                  "network":"Polkadot","id":"0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"
+                                }
+                              }
+                            }
+                          }
+                      - type: XcmVersionedMultiAssets
+                        value: {
+                          "V1":[
```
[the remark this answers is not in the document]
bernardoaraujor: I'm not sure I understand the suggestion. Doing the following breaks the tests:
```
-                          "V1":[
+                          V1:[
```

primitives/xcm/src/lib.rs
bernardoaraujor: https://github.com/paritytech/trappist/pull/109/commits/6d8df2b6377af68b4fbd2a9f32d3230ada1c8235

integration-tests/asset-trap/1_trap_claim_native_asset.yml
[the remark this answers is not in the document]
evilrobot-01: It breaks as it needs a space after the colon. All the other tests don't have properties wrapped in quotes, so it's standardising.

Happy to send a PR to your branch if you'd like.

evilrobot-01 (approved):
LGTM!

PR #110, opened 2023-01-13
fix: integration test formatting

evilrobot-01:
The latest release version of the `parachains-integration-tests` tool (v2.1) added yaml test file format validation (enforcement).


evilrobot-01:
Seems to be issues with CI downloading ubuntu packages, probably due to some network outage. Will try again later...

bernardoaraujor (approved):

PR #139, opened 2023-04-04
[ci] GHA to publish docker image

alvicsam:
PR adds github action that builds and publishes docker image on release

cc https://github.com/paritytech/ci_cd/issues/769

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           images: paritytech/trappist
 
-      - name: Build and push Docker image
+      - name: Build Docker image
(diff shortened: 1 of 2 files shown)


alvicsam:
Just noticed that the repo doesn't have Dockerfile :/

hbulgarini:
> Just noticed that the repo doesn't have Dockerfile :/

hey! we do, but it is not merged yet :P

https://github.com/paritytech/trappist/tree/hb-deploy-preparation/docker

I would like to take the opportunity to ask you if the docker image for the collator looks correct since you are the expert in this field. If it does, i can progress to merge that today.

sergejparity (approved):

Issue #137, opened 2023-02-28
[Trappist on Rococo] Remove `pallet-contracts-xcm` pallet


hbulgarini:
Before deploying Trappist on Rococo i think we should remove the `pallet-contracts-xcm` from the runtime and remove the chain extension for XCM on the `pallet_contracts::Config`.  The reason behind this is because the purpose of the `pallet-contracts-xcm` was to provide a PoC rather than a baseline for a production ready solution. As after a couple of months, nothing was done on this pallet besides our upgrades for Trappist making me think that the project/approach is kind of discarded for now. 
Also the approach about how to integrate XCM with the pallet_contracts + ink! is still under discussion and it's not even clear if chain extensions are going to be the final design decision for this purpose. 

CC: @stiiifff

hbulgarini:
@stiiifff i have moved this issue to M1 since i think it is cleaner to deploy to Rococo without this chain extension configured.

hbulgarini:
Pallet removed.

PR #149, opened 2023-04-14
Remove pallet-contracts-xcm pallet

hbulgarini:
Before deploying Trappist on Rococo i think we should remove the pallet-contracts-xcm from the runtime and remove the chain extension for XCM on the pallet_contracts::Config. The reason behind this is because the purpose of the pallet-contracts-xcm was to provide a PoC rather than a baseline for a production ready solution. As after a couple of months, nothing was done on this pallet besides our upgrades for Trappist making me think that the project/approach is kind of discarded for now.
Also the approach about how to integrate XCM with the pallet_contracts + ink! is still under discussion and it's not even clear if chain extensions are going to be the final design decision for this purpose.

Issue addressed in this PR:
[[Trappist on Rococo] Remove pallet-contracts-xcm pallet#137](https://github.com/paritytech/trappist/issues/137)


stiiifff:
@hbulgarini Looks like it's also in the `stout` parachain runtime .. should we remove it from there as well ?

stiiifff (dismissed):

PR #148, opened 2023-04-14
Live chain spec structure for deployment.

hbulgarini:
# Live chain spec structure for deployment

This PR includes the Live chain spec structure for deploying Trappist on Rococo.
Some of the decisions that are being included as part of this change:

## Parachain ID
1836

## Pre funded accounts

- 6 pre-funded accounts with 1.5 M of HOP.
- A multisig account with the 6 pre-funded accounts, with a 2 out of 6 threshold. This multisig was pre-founded with 1M of HOP.
- The multisig account was granted with Root origin through the sudo pallet.

## Council

- Initialize council members to the 6 pre-funded accounts
- Increases the `MotionDuration` for 48 hours/

## Considerations

This is the base for the chain spec with some of the design decisions already included. New configuration will be added as part of other deployment related tasks.

This PR address the issues:
closes [[Trappist on Rococo] Configure & initialize sudo account with multisig#143](https://github.com/paritytech/trappist/issues/143)
closes [[Trappist on Rococo] Configure Council#142](https://github.com/paritytech/trappist/issues/142)

PR #153, opened 2023-04-17
Add Treasury pallet to Trappist.

### LINKED ISSUE #145: [Trappist on Rococo] Add and configure Treasury pallet
- Add Treasury pallet to Trappist's runtime
- Configure Treasury pallet, see [Kusama's Treasury config](https://github.com/paritytech/polkadot/blob/b30c015050373ace42236df8d74f7bbf10bdbb42/runtime/kusama/src/lib.rs#L602-L638) as an example.
- Configure Transaction Payment pallet to transfer 80% of tx fees to the Treasury, and 20% to the block author. See Kusama & Polkadot's [DealWithFees implementation](https://github.com/paritytech/polkadot/blob/b30c015050373ace42236df8d74f7bbf10bdbb42/runtime/common/src/impls.rs#L39-L60) as an example.

hbulgarini:
This PR is configuring the treasury pallet according to the requirements described on the issue #145.

Most of the values have been taken from the Kusama treasury configuration. But the following configs were set empty:

```rust
	type Burn = ();
	type BurnDestination = ();
	type SpendFunds = ();
```

Burn: For the moment we did agree any burning mechanism to provoke deflation over the supply, so there no funds will be burnt.
SpendsFunds: This is mostly used for the Bounty pallet, that it was not decided to be implemented at this time.

Closes #145 
Depends on #148


hbulgarini:
Canceled.

PR #158, opened 2023-04-20
Fix stout compilation error.

hbulgarini:
This PR addresses the compilation issue created in the PR #148.

It also adds some enhancements as creating the `stout-collator` instead of having it to create it manually.

NOTE: For some reason, zombienet is not able to generate and parse the chain specs, but it is working when i generate it manually. I suspect something is happening with zombienet (i have heard others having different issues)

Closes #156

PR #159, opened 2023-04-24
Added Democracy pallet

### LINKED ISSUE #152: [Trappist on Rococo] Add democracy pallet to Rococo.
If we want to replicate a more real production experience when managing the live version of Trappist, we need to configure the democracy pallet like for example executing runtime upgrades.

<Specs and more details are still to be discussed and defined>

valentinfernandez1:
This PR includes the democracy pallet into the Trappist runtime as part of the  issue #152.

Most value are based of the Rococo config for this pallet. The Periods can be discussed depending on the objective of the project and how is going to be intended to be used by other teams.

```rust
type Slash = (); 
```
Slash was defined empty for the same reason. Once we define how different teams are intended to interact with the pallet this can be modified.

Closes #152

### DIFF
--- a/runtime/trappist/src/lib.rs
+++ b/runtime/trappist/src/lib.rs
@@ -25,7 +25,6 @@ pub mod constants;
 mod contracts;
 pub mod xcm_config;
 
-use common::DAYS;
 use cumulus_pallet_parachain_system::RelayNumberStrictlyIncreases;
 use sp_api::impl_runtime_apis;
 use sp_core::{crypto::KeyTypeId, ConstU8, OpaqueMetadata};
@@ -64,7 +63,7 @@ pub use parachains_common as common;
 pub use parachains_common::{
 	impls::{AssetsToBlockAuthor, DealWithFees},
 	opaque, AccountId, AssetId, AuraId, Balance, BlockNumber, Hash, Header, Index, Signature,
-	AVERAGE_ON_INITIALIZE_RATIO, HOURS, MAXIMUM_BLOCK_WEIGHT, MINUTES, NORMAL_DISPATCH_RATIO,
+	AVERAGE_ON_INITIALIZE_RATIO, DAYS, HOURS, MAXIMUM_BLOCK_WEIGHT, MINUTES, NORMAL_DISPATCH_RATIO,
 	SLOT_DURATION,
 };
 use xcm_config::{CollatorSelectionUpdateOrigin, RelayLocation};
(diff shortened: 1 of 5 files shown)


runtime/trappist/src/lib.rs
```diff
@@ -25,6 +25,7 @@ pub mod constants;
 mod contracts;
 pub mod xcm_config;
 
+use common::DAYS;
```
hbulgarini, line 28: Minor. For consistency i would import `DAYS` from here:
https://github.com/paritytech/trappist/pull/159/files#diff-1f7754547bbff248c6d0c46b32d579f4ecd082d31b014b6d141d116d6f2e0df0R64

runtime/trappist/src/lib.rs
```diff
@@ -478,6 +479,62 @@ impl pallet_preimage::Config for Runtime {
...
+	pub const MinimumDeposit: Balance = 100 * CENTS;
+	pub const MaxVotes: u32 = 100;
+	pub const MaxProposals: u32 = 100;
+	pub const InstantAllowed: bool = true;
+}
+
+impl pallet_democracy::Config for Runtime {
+	type WeightInfo = pallet_democracy::weights::SubstrateWeight<Runtime>;
+	type RuntimeEvent = RuntimeEvent;
+	type Scheduler = Scheduler; //<<check
```
hbulgarini, line 498: What needs to be checked here?

runtime/trappist/src/lib.rs
```diff
@@ -478,6 +479,62 @@ impl pallet_preimage::Config for Runtime {
...
+	type Preimages = Preimage;
+	type Currency = Balances;
+	type MinimumDeposit = MinimumDeposit;
+	type InstantAllowed = InstantAllowed;
+	type MaxVotes = MaxVotes;
+	type MaxProposals = MaxProposals;
+	type MaxDeposits = ConstU32<100>;
+	type MaxBlacklisted = ConstU32<100>;
+	type Slash = ();  //Define
```
hbulgarini, line 507: This can be kept empty until we add the Treasury pallet and we point the slashes to there.

hbulgarini (commented):
I have left some comments & questions.

I would wait for this PR to be merged until #158 is merged though.

runtime/trappist/src/lib.rs
valentinfernandez1: My bad i forgot to delete that comment, I was getting an error with rust-analyzer

stiiifff (approved):

PR #160, opened 2023-04-26
Add Treasury pallet to Trappist.

hbulgarini:
This PR is configuring the treasury pallet according to the requirements described on the issue https://github.com/paritytech/trappist/issues/145.

Most of the values have been taken from the Kusama treasury configuration. But the following configs were set empty:

```
	type Burn = ();
	type BurnDestination = ();
	type SpendFunds = ();
```

Burn: For the moment we didn't agree any burning mechanism to provoke deflation over the supply, so no funds will be burnt.
SpendsFunds: This is mostly used for the Bounty pallet, that it was not decided to be implemented at this time.

Closes https://github.com/paritytech/trappist/issues/145


runtime/trappist-common/src/impls.rs
```diff
@@ -0,0 +1,326 @@
...
+			use pallet_treasury::Pallet as Treasury;
+			<Treasury<R> as OnUnbalanced<_>>::on_unbalanced(split.0);
+			<ToAuthor<R> as OnUnbalanced<_>>::on_unbalanced(split.1);
+		}
+	}
+}
+
+/// A `HandleCredit` implementation that naively transfers the fees to the block author.
+/// Will drop and burn the assets in case the transfer fails.
+pub struct AssetsToBlockAuthor<R>(PhantomData<R>);
```
hbulgarini, line 78: These were taken from Statemine. At the moment they are not used but they look interesting and we can consider in future iterations to use them.

runtime/trappist-common/src/xcm_config.rs
```diff
@@ -0,0 +1,121 @@
+use crate::impls::AccountIdOf;
```
hbulgarini, line 1: At the moment we have https://github.com/paritytech/trappist/tree/main/primitives/xcm which seems to fill the same need. I wonder if we should move these to the primitives folder or if primitives should be moved here. This convention comes from Statemine as well.

runtime/trappist/src/lib.rs
```diff
@@ -502,7 +505,7 @@ impl pallet_democracy::Config for Runtime {
 	type MaxProposals = MaxProposals;
 	type MaxDeposits = ConstU32<100>;
 	type MaxBlacklisted = ConstU32<100>;
-	type Slash = ();  
+	type Slash = Treasury;
```
hbulgarini, line 508: CC: @valentinfernandez1

valentinfernandez1:
I'm not completely sure I understand the purpose of the `trappist-common` runtime.

valentinfernandez1 (approved):

runtime/trappist-common/src/impls.rs
stiiifff: I wondered why the code was copied ? ... instead of referencing existing crates (e.g. parachain-commons)

hbulgarini: I like having them in our code so we keep the flexibility to perform any custom logic and do not depend on Statemine at the time to support this impls. What do you think? do you think that is more beneficial to referencing the existing crates?

stiiifff: Yes, if there's no customization yet, I find it overkill to have a `trappist-common` crate. We might as well reference `parachains-common`, that makes the code lineage very clear (i.e. where is this logic (= trait implementations) coming from). If `AssetsToBlockAuthor` (which is coming from Statemine) looks interesting but is unused for now, just take a note of it, but don't include it in the code.

hbulgarini: I will remove the traits `AssetsToBlockAuthor`, `NonZeroIssuance` and `AssetsFrom` as they are not used now. However i realized that we still have to keep the custom implementation of `DealWithFees` in the `trappist-common` crate since the `DealWithFees` from `parachains-common` does not split the 80/20 of the fees as we agreed.

stiiifff: Put it in the Trappist runtime's for now (in an `impls.rs' file). Let's keep things simple.

hbulgarini: Addressed.

runtime/trappist/src/lib.rs
```diff
@@ -616,18 +662,21 @@ construct_runtime!(
 
 		// Runtime features
 		Contracts: pallet_contracts = 40,
-		Council: pallet_collective::<Instance1> = 41,
 		Assets: pallet_assets = 42,
```
stiiifff, line 665: No more 41 ... 🥲️

runtime/trappist/src/lib.rs
```diff
@@ -616,18 +662,21 @@ construct_runtime!(
...
 		Scheduler: pallet_scheduler = 45,
 		Preimage: pallet_preimage = 46,
-		Democracy: pallet_democracy = 47,
 
 		// Handy utilities.
 		Utility: pallet_utility::{Pallet, Call, Event} = 50,
 		Multisig: pallet_multisig::{Pallet, Call, Storage, Event<T>} = 51,
 
+		// Governance related
+		Council: pallet_collective::<Instance1> = 61,
```
stiiifff, line 676: Ok for grouping Governance-related pallets together ... but why not start at 60 🥲️ (like in other groups) .. let's be consistent.

hbulgarini:
@stiiifff instances number updated.

stiiifff (approved):

PR #163, opened 2023-04-27
Lockdown Mode pallet.

### LINKED ISSUE #144: [Trappist on Rococo] Add pallet for maintenance mode
Add a pallet to Trappist's runtime to allow for a maintenance (aka emergency) mode.
[Moonbeam](https://github.com/PureStake/moonbeam/tree/master/pallets/maintenance-mode) and [Acala](https://github.com/AcalaNetwork/Acala/tree/master/modules/transaction-pause), among others, have a similar feature.

hbulgarini:
This PR implements a lockdown mode pallet inspired in the Moonbeam maintenance pallet implementation.

Specs can be found here: https://github.com/paritytech/trappist/issues/144#issuecomment-1521927519

Closes #144


pallets/maintenance-mode/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+#[cfg(test)]
+mod tests;
+
+#[cfg(feature = "runtime-benchmarks")]
+mod benchmarking;
+pub mod weights;
+pub use weights::*;
+
+pub const ACTIVATE: bool = true;
+pub const DEACTIVATE: bool = false;
```
stiiifff, line 19: I would use either a verb with past tense, or an adjective .. but not an infinitive verb, e.g. Activated or Active.

pallets/maintenance-mode/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+		type MaintenanceModeOrigin: EnsureOrigin<Self::RuntimeOrigin>;
+		type FilteredCalls: Contains<Self::RuntimeCall>;
+		type MaintenanceDmpHandler: DmpMessageHandler;
+		type XcmExecutorManager: PauseXcmExecution;
+		type WeightInfo: WeightInfo;
+	}
+
+	#[pallet::storage]
+	pub type MaintenanceModeOnOff<T: Config> = StorageValue<_, bool, ValueQuery>;
```
stiiifff, line 62: Maybe something like `MaintenanceModeStatus` instead of `MaintenanceModeOnOff` ?

hbulgarini:
The PR is ready to be reviewed now. Still needs to be confirmed, but it seems the chess pallet is the responsable for the wasm exceed during execution.

pallets/maintenance-mode/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+
+			Self::deposit_event(Event::MaintenanceModeDeactivated);
+
+			Ok(())
+		}
+	}
+
+	impl<T: Config> Contains<T::RuntimeCall> for Pallet<T> {
+		fn contains(call: &T::RuntimeCall) -> bool {
+			log::info!("Pallet Contains: {:?}", call);
```
kalaninja, line 134: is it triggered on every tx validation? if so, should be "debug" or even "trace" level

pallets/maintenance-mode/src/tests.rs
```diff
@@ -0,0 +1,71 @@
...
+
+		assert_noop!(
+			MaintenanceMode::deactivate_maintenance_mode(RuntimeOrigin::root(),),
+			Error::<Test>::MaintenanceModeAlreadyDeactivated
+		);
+	});
+}
+
+#[test]
+fn call_not_filtered_in_maintenance_mode() {
```
kalaninja, line 42: this should be `call_filtered_in_maintenance_mode`.

you have whitelisted only `Balance`
```
pub struct RuntimeFilteredCalls;
impl Contains<RuntimeCall> for RuntimeFilteredCalls {
	fn contains(call: &RuntimeCall) -> bool {
		match call {
			RuntimeCall::Balance(_) => true,
			_ => false,
		}
	}
}
```

pallets/maintenance-mode/src/tests.rs
```diff
@@ -0,0 +1,71 @@
...
+	new_test_ext().execute_with(|| {
+		assert_ok!(MaintenanceMode::activate_maintenance_mode(RuntimeOrigin::root()));
+		let remark_call = RuntimeCall::Remark(RemarkCall::store { remark: vec![1, 2, 3] });
+		let result: bool = MaintenanceMode::contains(&remark_call);
+		assert!(!result);
+	});
+}
+
+#[test]
+fn call_filtered_in_maintenance_mode() {
```
kalaninja, line 52: this is NOT filtered

pallets/maintenance-mode/src/tests.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+#[test]
+fn call_not_filtered_in_normal_mode() {
+	new_test_ext().execute_with(|| {
+		let maintenance_mode = MaintenanceModeStatus::<Test>::get();
+		assert_eq!(maintenance_mode, DEACTIVATEDD);
+		let balance_call = RuntimeCall::Balance(BalancesCall::transfer { dest: 1, value: 2 });
+		let result: bool = MaintenanceMode::contains(&balance_call);
+		assert!(!result);
```
kalaninja, line 69: should always be `true` in normal mode

pallets/maintenance-mode/src/weights.rs
```diff
@@ -0,0 +1,76 @@
...
+	fn activate_maintenance_mode() -> Weight {
+		Weight::from_ref_time(18_710_000)
+			.saturating_add(T::DbWeight::get().reads(2))
+			.saturating_add(T::DbWeight::get().writes(2))
+	}
+	// Storage: AssetRegistry AssetIdMultiLocation (r:1 w:1)
+	// Storage: AssetRegistry AssetMultiLocationId (r:0 w:1)
+	fn deactivate_maintenance_mode() -> Weight {
+		Weight::from_ref_time(16_570_000)
+			.saturating_add(T::DbWeight::get().reads(1))
```
kalaninja, line 55: Looks suspicious that 2 calls with such a similar behavior have different number of reads.

pallets/maintenance-mode/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+			T::MaintenanceModeOrigin::ensure_origin(origin)?;
+
+			ensure!(
+				!MaintenanceModeStatus::<T>::get(),
+				Error::<T>::MaintenanceModeAlreadyActivated
+			);
+
+			MaintenanceModeStatus::<T>::put(ACTIVATED);
+
+			if let Err(error) = T::XcmExecutorManager::suspend_xcm_execution() {
```
kalaninja, line 101: this error might be caused if the origin used is not allowed to resume or suspend the XCMP queue. I am not sure how much sense does it make to emit an event here, but I think it's worth putting it in a log as might indicate a configuration error.

kalaninja (changes requested):

pallets/maintenance-mode/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+
+	#[derive(Default)]
+	#[pallet::genesis_config]
+	/// Genesis config for maintenance mode pallet
+	pub struct GenesisConfig {}
+
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig {
+		fn build(&self) {
+			MaintenanceModeStatus::<T>::put(ACTIVATED);
```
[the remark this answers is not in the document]
hbulgarini: so we can have the chance to check that everything is fine before making the operations available.

pallets/maintenance-mode/src/lib.rs
hbulgarini: So the reason why the event is there is to inform the user that the maintenance mode was activated but it did not work for the DmpQueue. An error would revert the whole call (because the transactional macro) and in case of emergencies is better to have the extrinsic filtered. On the other hand, creating only a log might pass unperceived leaving the user to think that the maintenance was activated for both.

pallets/maintenance-mode/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+		type MaintenanceModeOrigin: EnsureOrigin<Self::RuntimeOrigin>;
+		type FilteredCalls: Contains<Self::RuntimeCall>;
+		type MaintenanceDmpHandler: DmpMessageHandler;
+		type XcmExecutorManager: PauseXcmExecution;
+		type WeightInfo: WeightInfo;
+	}
+
+	#[pallet::storage]
+	pub type MaintenanceModeStatus<T: Config> = StorageValue<_, bool, ValueQuery>;
```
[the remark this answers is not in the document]
hbulgarini: is a valid comment, however my personal preference is to keep using the `::get()` instead of the getter. I think that 
`MaintenanceModeStatus::<T>::get()` gives more context than `Self::is_active()`. Also i have seen a lot of getters not used in many pallets.

stiiifff:
@hbulgarini I think calling this pallet `lockdown-mode` would better represent the intent (I just found out that Apple has such a [mode](https://support.apple.com/en-us/HT212650) in iOS).

pallets/maintenance-mode/src/lib.rs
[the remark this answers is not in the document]
valentinfernandez1: I feel like both approaches are good for different reasons. `Self::is_active()` is cleaner  in terms of code ergonomics and you can automatically assume that it will return a `boolean`, but in order to understand it you either have to read the hole line to get extra context or go to the definition of the function.

Whereas `MaintenanceModeStatus::<T>::get()` is pretty self explainatory, but not as clean.

I think is certainly a matter of taste here.

pallets/maintenance-mode/src/weights.rs
hbulgarini: I forgot running the weights and what we had here was just a placeholder. Should be good now.

hbulgarini:
@stiiifff @kalaninja @valentinfernandez1 , i have addressed all the comments.

pallets/maintenance-mode/src/lib.rs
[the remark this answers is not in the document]
kalaninja: I got it, but it doesn't seem to be a responsibility of a pallet. This decision (on or off at the start) should be taken on a composition root level. My suggestion here is to add `activated` into the genesis config
```
	#[pallet::genesis_config]
	/// Genesis config for lockdown mode pallet
	pub struct GenesisConfig<T: Config> {
		pub activated: bool,
		pub _phantom: PhantomData<T>,
	}

	impl<T: Config> Default for GenesisConfig<T> {
		fn default() -> Self {
			Self { activated: ACTIVATED, _phantom: PhantomData }
		}
	}

	#[pallet::genesis_build]
	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
		fn build(&self) {
			LockdownModeStatus::<T>::put(&self.activated);
		}
	}
```

this will also let you to set initial state for you tests
```
pub fn new_test_ext(activated: bool) -> sp_io::TestExternalities {
	let mut storage = system::GenesisConfig::default().build_storage::<Test>().unwrap();
	pallet_lockdown_mode::GenesisConfig::<Test> { activated, _phantom: PhantomData }
		.assimilate_storage(&mut storage)
		.unwrap();

	storage.into()
}
```

So that you don't need to activate lockdown mode first to deactivate it in the next step.

You can also add a test to make sure that the storage is initialized properly
```
#[test]
fn initialized() {
	[true, false].iter().for_each(|expected| {
		new_test_ext(*expected).execute_with(|| {
			let lockdown_mode = LockdownModeStatus::<Test>::get();
			assert_eq!(lockdown_mode, *expected);
		});
	});
}
```

pallets/lockdown-mode/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+		fn build(&self) {
+			LockdownModeStatus::<T>::put(ACTIVATED);
+		}
+	}
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+		type LockdownModeOrigin: EnsureOrigin<Self::RuntimeOrigin>;
+		type FilteredCalls: Contains<Self::RuntimeCall>;
```
kalaninja, line 54: I think this name is confusing, it is hard to understand is it whitelisting or blacklisting. how about `WhiteListedCalls`?

pallets/lockdown-mode/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+		pub fn activate_lockdown_mode(origin: OriginFor<T>) -> DispatchResult {
+			T::LockdownModeOrigin::ensure_origin(origin)?;
+
+			ensure!(!LockdownModeStatus::<T>::get(), Error::<T>::LockdownModeAlreadyActivated);
+
+			LockdownModeStatus::<T>::put(ACTIVATED);
+
+			if let Err(error) = T::XcmExecutorManager::suspend_xcm_execution() {
+				log::error!("Failed to suspend idle XCM execution {:?}", error);
+				<Pallet<T>>::deposit_event(Event::FailedToSuspendIdleXcmExecution { error });
```
kalaninja, line 99: `<Pallet<T>>::deposit_event` just to make it look the same everywhere

pallets/lockdown-mode/src/lib.rs
hbulgarini: I take that, i think the purpose is to  black list , so it would be better: `BlackListedCalls`

pallets/lockdown-mode/src/lib.rs
hbulgarini: i don't get your point in this one, could you please elaborate?

kalaninja: oops, I meant `Self::deposit_event`

stiiifff (approved):

pallets/maintenance-mode/src/lib.rs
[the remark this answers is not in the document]
stiiifff: @hbulgarini I think @kalaninja makes sense. If you can apply those changes, we should be good then.

hbulgarini:
@kalaninja , i have to solve some conflicts on this branch because of the merge of the PR #165, could you please make sure that something was not incorrectly solved during the sync?  Then if everything looks fine, you can just approve and merge. @stiiifff already approved it but due to the re-sync the approval was missed.

stiiifff:
@hbulgarini approved but I think it'd be nice to implement @kalaninja 's suggestion ([see above](https://github.com/paritytech/trappist/pull/163#discussion_r1192183216)).

kalaninja (approved):

stiiifff (approved):
LGTM. Let's create a new issue with @kalaninja's improvement suggestion.

pallets/maintenance-mode/src/lib.rs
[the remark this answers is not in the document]
hbulgarini: https://github.com/paritytech/trappist/issues/172

PR #173, opened 2023-05-12
`asset-registry` refactoring

kalaninja:
- Fix benchmarks. Benchmarks fail because of this check `ensure!(T::Assets::asset_exists(asset_id), Error::<T>::AssetDoesNotExist)`. The asset being used has to exist in the pallet responsible for storing assets (pallet_assets in our case)
- Refactor tests. Increased atomicity.

### DIFF
--- a/pallets/asset-registry/src/lib.rs
+++ b/pallets/asset-registry/src/lib.rs
@@ -107,8 +107,9 @@ pub mod pallet {
 				Error::<T>::WrongMultiLocation
 			);
 
-			// register asset
+			// register asset_id => asset_multi_location
 			AssetIdMultiLocation::<T>::insert(asset_id, &asset_multi_location);
+			// register asset_multi_location => asset_id
 			AssetMultiLocationId::<T>::insert(&asset_multi_location, asset_id);
 
 			Self::deposit_event(Event::ReserveAssetRegistered { asset_id, asset_multi_location });
@@ -123,10 +124,11 @@ pub mod pallet {
 		) -> DispatchResult {
 			T::ReserveAssetModifierOrigin::ensure_origin(origin)?;
 
-			// unregister asset
+			// remove asset_id => asset_multi_location, while getting the value
 			let asset_multi_location =
 				AssetIdMultiLocation::<T>::mutate_exists(asset_id, Option::take)
 					.ok_or(Error::<T>::AssetIsNotRegistered)?;
+			// remove asset_multi_location => asset_id
 			AssetMultiLocationId::<T>::remove(&asset_multi_location);
 
 			Self::deposit_event(Event::ReserveAssetUnregistered { asset_id, asset_multi_location });
--- a/pallets/asset-registry/src/lib.rs
+++ b/pallets/asset-registry/src/lib.rs
@@ -33,21 +33,27 @@ pub mod pallet {
 	type AssetIdOf<T> =
 		<<T as Config>::Assets as Inspect<<T as frame_system::Config>::AccountId>>::AssetId;
 
+	#[cfg(feature = "runtime-benchmarks")]
+	pub trait BenchmarkHelper<AssetId> {
+		fn get_registered_asset() -> AssetId;
+	}
+
 	#[pallet::config]
 	pub trait Config: frame_system::Config {
 		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
 		type ReserveAssetModifierOrigin: EnsureOrigin<Self::RuntimeOrigin>;
 		type Assets: Inspect<Self::AccountId>;
 		type WeightInfo: WeightInfo;
+		/// Helper trait for benchmarks.
+		#[cfg(feature = "runtime-benchmarks")]
+		type BenchmarkHelper: BenchmarkHelper<AssetIdOf<Self>>;
 	}
 
 	#[pallet::storage]
-	#[pallet::getter(fn asset_id_multilocation)]
 	pub type AssetIdMultiLocation<T: Config> =
 		StorageMap<_, Blake2_128Concat, AssetIdOf<T>, MultiLocation>;
 
 	#[pallet::storage]
-	#[pallet::getter(fn asset_multilocation_id)]
 	pub type AssetMultiLocationId<T: Config> =
 		StorageMap<_, Blake2_128Concat, MultiLocation, AssetIdOf<T>>;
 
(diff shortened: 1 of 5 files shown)


runtime/trappist/src/lib.rs
```diff
@@ -559,11 +559,30 @@ impl pallet_dex::Config for Runtime {
...
+pub struct AssetRegistryBenchmarkHelper;
+#[cfg(feature = "runtime-benchmarks")]
+impl pallet_asset_registry::BenchmarkHelper<AssetId> for AssetRegistryBenchmarkHelper {
+	fn get_registered_asset() -> AssetId {
+		use sp_runtime::traits::StaticLookup;
+
+		let root = frame_system::RawOrigin::Root.into();
+		let asset_id = 1;
+		let caller = frame_benchmarking::whitelisted_caller();
+		let caller_lookup = <Runtime as frame_system::Config>::Lookup::unlookup(caller);
```
hbulgarini, line 572: Is the whole point of the helper to create an asset with the `frame_benchmarking::whitelisted_caller()` ?

hbulgarini (commented):
I dropped two questions that are not blockers for merging this PR. However i think it would be nice to include a little bit more details in the PR description documenting how this is needed and why the current benchmark process is not working as expected. At least on my side i don't have 100% clear why the helper is needed instead of tweaking the GenesisConfig of the mocked runtime.

pallets/asset-registry/src/lib.rs
```diff
@@ -119,12 +123,10 @@ pub mod pallet {
...
 			T::ReserveAssetModifierOrigin::ensure_origin(origin)?;
 
-			// verify asset is registered
-			let asset_multi_location =
-				AssetIdMultiLocation::<T>::get(asset_id).ok_or(Error::<T>::AssetIsNotRegistered)?;
-
 			// unregister asset
-			AssetIdMultiLocation::<T>::remove(asset_id);
+			let asset_multi_location =
+				AssetIdMultiLocation::<T>::mutate_exists(asset_id, Option::take)
```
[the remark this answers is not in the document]
kalaninja: the idea here is that `mutate_exists(id, Option::take)` does get + remove with a single operation.
But readability might be a bit worse, so I am not sure myself.

runtime/trappist/src/lib.rs
kalaninja: I put a description in the comment. But, the main point is to create an asset in `pallet_assets` and return it to `asset-registry`, because the registry cannot do it on its own.

hbulgarini:
> > I dropped two questions that are not blockers for merging this PR. However i think it would be nice to include a little bit more details in the PR description documenting how this is needed and why the current benchmark process is not working as expected. At least on my side i don't have 100% clear why the helper is needed instead of tweaking the GenesisConfig of the mocked runtime.
> 
> We cannot use mocks in benchmarks. Pallets are implemented abstracting lots of things. Exact implementations of this abstractions affect the real performance. So benchmarks are executed on real runtimes and their results might differ for different runtimes, depending on how you parameterize it. So benchmarks have to be runtime agnostic, because their pallet can be used multiple ways and multiple configurations. Which is a problem in case of pallet coupling.
> 
> Since `asset-registry` is dependent on abstraction over `Assets` it has to prepare the stage for a benchmark in an abstract way too. `BenchmarkHelper` is a way for a pallet to say: I don't know how you manage assets in your runtime, but I need you to give me an Id of a registered asset for my benchmarks.

i get it know, i think the implementation looks all right. Thanks for the explanation.

hbulgarini (approved):

pallets/asset-registry/src/lib.rs
```diff
@@ -119,12 +123,10 @@ pub mod pallet {
...
+					.ok_or(Error::<T>::AssetIsNotRegistered)?;
 			AssetMultiLocationId::<T>::remove(&asset_multi_location);
```
hbulgarini, line 130: is it possible that you forgot to remove this one taking into account that `mutate_exists(id, Option::take)` is already removing it?

kalaninja: No, there are 2 storages `AssetIdMultiLocation` and `AssetMultiLocationId`. The first is a map `id => MultiLocation` and the second is `MultiLocation => id`. The logic needs to get `MultiLocation` out of the 1st storage while removing it, and remove this `MultiLocation` from the second storage afterwards. So the second remove is just remove.
Initially, I implemented it as a functor:
```
pub fn unregister_reserve_asset(
	origin: OriginFor<T>,
	asset_id: AssetIdOf<T>,
) -> DispatchResult {
	T::ReserveAssetModifierOrigin::ensure_origin(origin)?;

	// unregister asset
	AssetIdMultiLocation::<T>::mutate_exists(asset_id, Option::take)
		.map(|asset_multi_location| {
			AssetMultiLocationId::<T>::remove(&asset_multi_location);
			Self::deposit_event(Event::ReserveAssetUnregistered {
				asset_id,
				asset_multi_location,
			});
		})
		.ok_or(Error::<T>::AssetIsNotRegistered.into())
}
```
But thought that the readability is even worse. And nesting grows. But might be an option if you like it. Or I can return back to how it was. Mainly is a question of taste.

hbulgarini: oh! i didn't noticed the difference about those two .. i read them very quickly and i thought you forgot to remove that. In terms of readability i agree with you that it does not add value honestly, but i think this is a very good example for an alternative removal than `get` + `remove`. So my vote goes for keeping it but adding a comment specifying exactly what you have described.

hbulgarini (approved):

pallets/asset-registry/src/lib.rs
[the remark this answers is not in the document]
stiiifff: As @hbulgarini mentioned, I think a comment will do.

stiiifff (approved):

pallets/asset-registry/src/lib.rs
kalaninja: done

pallets/asset-registry/src/lib.rs
```diff
@@ -119,12 +124,11 @@ pub mod pallet {
...
 			T::ReserveAssetModifierOrigin::ensure_origin(origin)?;
 
-			// verify asset is registered
+			// remove asset_id => asset_multi_location, while getting the value
 			let asset_multi_location =
-				AssetIdMultiLocation::<T>::get(asset_id).ok_or(Error::<T>::AssetIsNotRegistered)?;
-
-			// unregister asset
-			AssetIdMultiLocation::<T>::remove(asset_id);
+				AssetIdMultiLocation::<T>::mutate_exists(asset_id, Option::take)
```
hbulgarini, line 129:
```suggestion
                 // mutate_exists(id, Option::take) reads the value and delete it in one operation. Even if it's not as clear as `get` + `remove`, it is an alternative pattern when removing items.
				AssetIdMultiLocation::<T>::mutate_exists(asset_id, Option::take)
```

stiiifff: I think it's good enough, people can also read the doc for the method 😉️

PR #181, opened 2023-05-22
Lockdown DmpMessageHandler fix

valentinfernandez1:
Fixes a previous error in the Lockdown `DmpMessageHandler` that would cause the chain to execute an Unreachable WASM instruction preventing the chain from producing blocks.

This was caused after replacing `DmpQueue` for `LockdownMode`in `cumulus_pallet_parachain_system`
```rust
impl cumulus_pallet_parachain_system::Config for Runtime {
	....
	type DmpMessageHandler = LockdownMode;
	....
}
```


hbulgarini (approved):
lgtm!

PR #184, opened 2023-05-22
Backport PRs from main branch

valentinfernandez1:
It backports the changes introduced to the main branch and updates them to v9.40. The PRs updated are:
+ Code cleanup #166 
+ Benchmark drop_assets() of TrappistDropAssets #113 
+ Lockdown Mode pallet #163 
+ Asset-registry refactoring #173 
+ CI refactoring #174 

The weights of the pallets were re-benchmarked to use Weights V2. 

---
It also fixes Trappist and Stout `UnitWeightCost` to match the config of  the [parachain-template](https://github.com/paritytech/cumulus/blob/e05c8d7f71734ed71188337c6cb0d30715f6320f/parachain-template/runtime/src/xcm_config.rs#L80).
```rust

parameter_types! {
	....
	pub UnitWeightCost: Weight = Weight::from_parts(1_000_000_000, 64 * 1024);
	....
}
```


stiiifff (changes requested):
This is very difficult to review as-is. Having all PRs mixed up into one means the reviewer has to go back to the original PRs and do a very thorough job of re-examining everything.
My bad, maybe I should have mentioned it in the issue #184.
I would suggest to close it, and restart by opening one separate PR per the original PRs being back-ported. This will be much faster and less error-prone.

valentinfernandez1:
You are right, I didn't realized how difficult it was going to be to review so many different changes. I will split them into different PRs to simplify it. 

Also it makes more sense this way since more things will need to be backported like #185 and this way it's going to be easier to identify each change.

Issue #199, opened 2023-05-29
[0.9.40] Add `query_account_balances` runtime api for pallet assets.


hbulgarini:
During release 0.9.40, the `query_account_balances` runtime api for querying the assets balances were implemented.

We can have the details here:
https://github.com/paritytech/substrate/pull/13352

This is what it has to be added in the runtime (please check the [assets_common](https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/assets/common/src/runtime_api.rs#L32)):
```rust
	impl assets_common::runtime_api::FungiblesApi<
		Block,
		AccountId,
	> for Runtime
	{
		fn query_account_balances(account: AccountId) -> Result<xcm::VersionedMultiAssets, assets_common::runtime_api::FungiblesAccessError> {
			use assets_common::fungible_conversion::{convert, convert_balance};
			Ok([
				// collect pallet_balance
				{
					let balance = Balances::free_balance(account.clone());
					if balance > 0 {
						vec![convert_balance::<KsmLocation, Balance>(balance)?]
					} else {
						vec![]
					}
				},
				// collect pallet_assets (TrustBackedAssets)
				convert::<_, _, _, _, TrustBackedAssetsConvertedConcreteId>(
					Assets::account_balances(account.clone())
						.iter()
						.filter(|(_, balance)| balance > &0)
				)?,
				// collect pallet_assets (ForeignAssets)
				convert::<_, _, _, _, ForeignAssetsConvertedConcreteId>(
					ForeignAssets::account_balances(account)
						.iter()
						.filter(|(_, balance)| balance > &0)
				)?,
				// collect ... e.g. other tokens
			].concat().into())
		}
	}
```

CC: @stiiifff @valentinfernandez1

valentinfernandez1:
I like this implementation. We would have to remove this part since we don't have `ForeignAssets`.
```rust
                                ....
                                // collect pallet_assets (ForeignAssets)
				convert::<_, _, _, _, ForeignAssetsConvertedConcreteId>(
					ForeignAssets::account_balances(account)
						.iter()
						.filter(|(_, balance)| balance > &0)
				)?,
```
But everything else looks good to me. I can add it today. 

In regards to your second comment, why do you consider that both `convert_ref` should coexist in the same file??

hbulgarini:
> In regards to your second comment, why do you consider that both convert_ref should coexist in the same file??

[...]

Issue #169, opened 2023-05-10
[0.9.40] Change `WeightToFee` struct to charge for proof size weight.


hbulgarini:
As per it was described in the following issues/PRs:

https://github.com/paritytech/cumulus/issues/2322
https://github.com/paritytech/substrate/pull/13612
https://github.com/paritytech/cumulus/pull/2326

We should add the `proof_size` calculation to the `WeightToFee` structure.

stiiifff:
@hbulgarini @valentinfernandez1 We can close this one, right ?

PR #202, opened 2023-05-29
Query account balances using `accounts_common`

valentinfernandez1:
Adds the `query_account_balance` functionality as suggested in issue #199 using the crate `accounts_common` from cumulus. this allows to query multiple balances from an account with a single runtime api call.

```rust
        impl assets_common::runtime_api::FungiblesApi<
		Block,
		AccountId,
	> for Runtime
	{
		fn query_account_balances(account: AccountId) -> Result<Vec<MultiAsset>, assets_common::runtime_api::FungiblesAccessError> {
			use assets_common::fungible_conversion::{convert, convert_balance};
			Ok([
				// collect pallet_balance
				{
					let balance = Balances::free_balance(account.clone());
					if balance > 0 {
						vec![convert_balance::<RelayLocation, Balance>(balance)?]
					} else {
						vec![]
					}
				},
				// collect pallet_assets (TrustBackedAssets)
				convert::<_, _, _, _, TrustBackedAssetsConvertedConcreteId>(
					Assets::account_balances(account.clone())
						.iter()
						.filter(|(_, balance)| balance > &0)
				)?
			].concat().into())
		}
	}
```
Also the Converter `TrustBackedAssetsConvertedConcreteId` was added

----
Inside `trappist/xcm_config.rs` LocalFungiblesTransactor was modified including this new Assetconverter from:
```rust
/// Means for transacting assets besides the native currency on this chain.
pub type LocalFungiblesTransactor = FungiblesAdapter<
	.....
	// Use this currency when it is a fungible asset matching the given location or name:
	ConvertedConcreteId<
		AssetIdForTrustBackedAssets,
		Balance,
		AsPrefixedGeneralIndex<AssetsPalletLocation, AssetIdForTrustBackedAssets, JustTry>,
		JustTry,
	>,
	....
>;
```
to:
```rust
/// Means for transacting assets besides the native currency on this chain.
pub type LocalFungiblesTransactor = FungiblesAdapter<
	.....
	// Use this currency when it is a fungible asset matching the given location or name:
	TrustBackedAssetsConvertedConcreteId,
	....
>;
```
Since this accomplishes the same thing and this converter will probably be needed for #169.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -407,6 +407,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 
 [[package]]
+name = "assets-common"
+version = "0.1.0"
+source = "git+https://github.com/paritytech/cumulus?branch=polkadot-v0.9.40#e05c8d7f71734ed71188337c6cb0d30715f6320f"
+dependencies = [
+ "frame-support",
+ "parachains-common",
+ "parity-scale-codec",
+ "sp-api",
+ "sp-std",
+ "substrate-wasm-builder",
+ "xcm",
+ "xcm-builder",
+ "xcm-executor",
+]
+
+[[package]]
 name = "async-channel"
 version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -12398,6 +12414,7 @@ dependencies = [
 name = "trappist-runtime"
 version = "1.0.0"
 dependencies = [
+ "assets-common",
  "cumulus-pallet-aura-ext",
  "cumulus-pallet-dmp-queue",
  "cumulus-pallet-parachain-system",
--- a/primitives/xcm/src/lib.rs
+++ b/primitives/xcm/src/lib.rs
@@ -6,7 +6,7 @@ use frame_support::{
 	weights::Weight,
 };
 use sp_runtime::DispatchResult;
-use sp_std::{borrow::Borrow, marker::PhantomData, vec::Vec};
+use sp_std::{borrow::Borrow, marker::PhantomData};
 use xcm::{
 	latest::{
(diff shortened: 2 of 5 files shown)


valentinfernandez1:
> In addition to this i have noticed that the assets-common, contains a more interesting [MultiAssetConverter/convert_ref](https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/assets/common/src/fungible_conversion.rs#L47) implementation that the one we have in our [xcm-primitives](https://github.com/paritytech/trappist/blob/main/primitives/xcm/src/lib.rs#L26).
> 
> I see a couple of improvements in the asset-common as:
> 
>     Better error management.
>     It returns the amount in addition to the asset id.
> 
> This should also be considered in the asset-registry pallet that we are implementing.

Also this needs to be implemented but I still have doubts about this:

> i see a period of time where both XCM primitives will cohexist until the [xcm-primitives](https://github.com/paritytech/trappist/blob/main/primitives/xcm/src/lib.rs#L26) is removed or replaced.

hbulgarini:
> > In addition to this i have noticed that the assets-common, contains a more interesting [MultiAssetConverter/convert_ref](https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/assets/common/src/fungible_conversion.rs#L47) implementation that the one we have in our [xcm-primitives](https://github.com/paritytech/trappist/blob/main/primitives/xcm/src/lib.rs#L26).
> > I see a couple of improvements in the asset-common as:
> > 
> > * Better error management.
> > * It returns the amount in addition to the asset id.
> > 
> > This should also be considered in the asset-registry pallet that we are implementing.
> 
> Also this needs to be implemented but I still have doubts about this:
> 
> > i see a period of time where both XCM primitives will cohexist until the [xcm-primitives](https://github.com/paritytech/trappist/blob/main/primitives/xcm/src/lib.rs#L26) is removed or replaced.

https://github.com/paritytech/trappist/issues/199#issuecomment-1567924699

valentinfernandez1:
I would resolve the `convert_ref` implementation and how it should be changed inside the `asset_registry` in a different PR.

PR #204, opened 2023-05-31
Fix XCM benchmarks

kalaninja:
Need to specify `SelfReserve` in `get_multi_asset`, because of the Matcher type in `LocalAssetTransactor`

```rust
pub type LocalAssetTransactor = CurrencyAdapter<
	// Use this currency:
	Balances,
	// Use this currency when it is a fungible asset matching the given location or name:
	IsConcrete<SelfReserve>, // <--- matcher
	// Convert an XCM MultiLocation into a local account id:
	LocationToAccountId,
	AccountId,
	// We don't track any teleports of `Balances`.
	(),
>;
```


hbulgarini (approved):
thanks!

PR #207, opened 2023-06-02
Trappist Workspace refactor.

### LINKED ISSUE #178: Create workspace for improving Cargo.toml experience.
At the moment we are hardcoding every version for all the crates used in Trappist. Many of them are repeated across different Cargo.toml files. The goal is to implement a workspace in the root so all the imports are taken from one single place instead of the current approach.

Depends on #177.

hbulgarini:
This PR does a complete refactor of all the `Cargo.toml` files to import the crates from the workspace and not from any single file. 

Please consider that `xcm-simulator` was excluded from this since at the moment it is excluded from the workspace. Honestly at this point in time that we have to migrate, setting the simulator in the workspace is not a priority.

Closes #178


valentinfernandez1 (approved):
LGTM

stiiifff (approved):

PR #215, opened 2023-06-06
[WIP] WeightToFee refactor to include ProofSize

valentinfernandez1:
This PR introduces an update to the WeightToFee struct, addressing the changes discussed in issue #169. The update aligns the struct with the latest version included in release v0.9.40, which incorporates the following trait:
```rust
pub trait WeightToFee {
/// The type that is returned as result from calculation.
    type Balance: BaseArithmetic + From<u32> + Copy + Unsigned;

    /// Calculates the fee from the passed `weight`.
    fn weight_to_fee(weight: &Weight) -> Self::Balance;
}
```
The implementation of this functionality draws heavily from the implementation provided by [statemine (now called asset-hub)](https://github.com/paritytech/cumulus/blob/525d8901459dac41f615b6f8e905c1f0eed41efd/parachains/runtimes/assets/asset-hub-polkadot/src/constants.rs#LL58C2-L69C3). However, the `FeePolynomial` struct and some Trait functions that are used in statemine implementation are not available for v0.9.40. To address this, a custom trait called WeightCoefficientCalc has been implemented, which facilitates the necessary calculations.
```rust
pub trait WeightCoefficientCalc<Balance> {
    fn saturating_eval(&self, result: Balance, x: Balance) -> Balance;
}

impl<Balance> WeightCoefficientCalc<Balance> for WeightToFeeCoefficient<Balance>
where
    Balance: BaseArithmetic + From<u32> + Copy + Unsigned + SaturatedConversion,
{
    fn saturating_eval(&self, mut result: Balance, x: Balance) -> Balance {
	let power = x.saturating_pow(self.degree.into());

	let frac = self.coeff_frac * power; // Overflow safe.
	let integer = self.coeff_integer.saturating_mul(power);
	// Do not add them together here to avoid an underflow.

	if self.negative {
		result = result.saturating_sub(frac);
		result = result.saturating_sub(integer);
	} else {
		result = result.saturating_add(frac);
		result = result.saturating_add(integer);
	}

	result
    }
}
```
This implementation can be easily switched to use FeePolynomial after upgrading the substrate version to v0.9.42 or above.

----
Note: I plan to include additional tests to ensure that this works as intended specially on write heavy extrinsics, but feel free to leave your feedback on this change in the mean time.


valentinfernandez1:
This issue is going to be paused until #217 is done so we can add proper testing. The other alternative is to merge this change and open a new issue for testing.

runtime/stout/src/constants.rs
```diff
@@ -71,14 +71,4 @@ pub mod fee {
 			}]
 		}
 	}
-
-	pub fn base_tx_fee() -> Balance {
```
hbulgarini, line 75: could you please rebase this branch? these should be back-in :)

runtime/stout/src/xcm_config.rs
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{constants::fee::default_fee_per_second, AllPalletsWithSystem};
```
hbulgarini, line 16: could you please rebase this branch? these should be back-in :)

runtime/stout/src/xcm_config.rs
```diff
@@ -192,11 +192,6 @@ parameter_types! {
 	// Statemine's Assets pallet index
 	pub StatemineAssetsPalletLocation: MultiLocation =
 		MultiLocation::new(1, X2(Parachain(1000), PalletInstance(50)));
-	pub XUsdPerSecond: (AssetId, u128, u128) = (
```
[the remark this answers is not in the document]
valentinfernandez1: Sure

runtime/trappist/src/constants.rs
```diff
@@ -58,8 +60,38 @@ pub mod fee {
...
+	/// TODO: Once the runtime is upgraded to polkadot v0.9.42 or above refactor this using
+	/// the FeePolynomial struct that already includes the methods to make this calculations
+	/// and remove the custom WeightCoefficientCalc inside ./trappist/src/impls.rs
 	pub struct WeightToFee;
-	impl WeightToFeePolynomial for WeightToFee {
+	impl frame_support::weights::WeightToFee for WeightToFee {
+		type Balance = Balance;
+
+		fn weight_to_fee(weight: &Weight) -> Self::Balance {
+			let ref_poly: smallvec::SmallVec<[WeightToFeeCoefficient<Balance>; 4]> =
```
kalaninja, line 72: Rust can infer the type properly here, so you don't need to specify any. Still if you want you can use `WeightToFeeCoefficients<Self::Balance>` I don't see why it should be known that this is a type alias for smallvec.

runtime/trappist/src/constants.rs
```diff
@@ -58,8 +60,38 @@ pub mod fee {
...
+				RefTimeToFee::polynomial();
+			let proof_poly: smallvec::SmallVec<[WeightToFeeCoefficient<Balance>; 4]> =
+				ProofSizeToFee::polynomial();
+
+			// Get fee amount from ref_time based on the RefTime polynomial
+			let ref_fee: Balance = ref_poly.iter().fold(0, |acc, term| {
+				term.saturating_eval(acc, Balance::saturated_from(weight.ref_time()))
```
kalaninja, line 79: `Balance::saturated_from(weight.ref_time()` can be moved out of the closure, so you don't execute the same  cast on every coefficient.

runtime/trappist/src/impls.rs
```diff
@@ -111,6 +112,34 @@ impl xcm_primitives::PauseXcmExecution for XcmExecutionManager {
...
+}
+
+impl<Balance> WeightCoefficientCalc<Balance> for WeightToFeeCoefficient<Balance>
+where
+	Balance: BaseArithmetic + From<u32> + Copy + Unsigned + SaturatedConversion,
+{
+	fn saturating_eval(&self, mut result: Balance, x: Balance) -> Balance {
+		let power = x.saturating_pow(self.degree.into());
+
+		let frac = self.coeff_frac * power; // Overflow safe.
```
kalaninja, line 127: I think you can be more specific here and say that it's "Overflow safe" because `coeff_frac` is strictly less than 1.

kalaninja (commented):
I've noticed a comment referencing kusama and took a closer look. Is it ok that we are using constants from `kusama_runtime_constants`? I think trappist should be referencing those of polkadot.

runtime/trappist/src/constants.rs
valentinfernandez1: You are right, this is directly taken from [Substrate - FeePolynomial](https://paritytech.github.io/substrate/master/src/sp_weights/lib.rs.html#184) adapting it to this version, but this should also be changed in substrate.

Good catch!

valentinfernandez1:
> I've noticed a comment referencing kusama and took a closer look. Is it ok that we are using constants from `kusama_runtime_constants`? I think trappist should be referencing those of polkadot.

This should be discussed since it's also part of trappist v0.9.37 I don't know the reason why it was added initially maybe @stiiifff  can clarify the reasoning behind it.

stiiifff:
> > I've noticed a comment referencing kusama and took a closer look. Is it ok that we are using constants from `kusama_runtime_constants`? I think trappist should be referencing those of polkadot.
> 
> This should be discussed since it's also part of trappist v0.9.37 I don't know the reason why it was added initially maybe @stiiifff can clarify the reasoning behind it.

@valentinfernandez1 @kalaninja The intention there was/is to use the same numbers as Kusama for things like existential deposit, deposit fees, and currency-related constants (i.e. the chain's native token, HOP, is mirroring the KSM token for decimals, total emission).

stiiifff:
Looks good, @valentinfernandez1 let's get the build fixed, and then LGTM.

PR #221, opened 2023-06-09
Fix xcm reserve transfer from Rockmine <> Trappist.

hbulgarini:
This PR fixes the reserve transfer asset from Rockmine to Trappist.

It adds the following Trader tuple:

```rust
pub type Trader = (
	FixedRateOfFungible<RUsdPerSecond, ()>,
	FixedRateOfFungible<RocPerSecond, ()>,
	UsingComponents<WeightToFee, SelfReserve, AccountId, Balances, ToAuthor<Runtime>>,
);
```

As you can see, also `Roc`s can be computed as a reference value of 1 ROC = 7 RUSD"
```rust
	pub RUsdPerSecond: (xcm::v1::AssetId, u128) = (
		MultiLocation::new(1, X3(Parachain(1000), PalletInstance(50), GeneralIndex(1984))).into(),
		default_fee_per_second() * 10
	);
	/// Roc = 7 RUSD
	pub RocPerSecond: (AssetId, u128) = (MultiLocation::parent().into(), default_fee_per_second() * 70);
```

Also there was an issue with the pallet-xcm-benchmarking that was setting the weight for `reserve_asset_deposited ` to `u64::MAX`. I suspect this is a hack for CGP to avoid reserve based transfers from other reserves.

### DIFF
--- a/runtime/trappist/src/constants.rs
+++ b/runtime/trappist/src/constants.rs
@@ -73,4 +73,14 @@ pub mod fee {
 			}]
 		}
 	}
+
+	pub fn base_tx_fee() -> Balance {
+		CENTS / 10
+	}
+
+	pub fn default_fee_per_second() -> u128 {
+		let base_weight = Balance::from(ExtrinsicBaseWeight::get().ref_time());
+		let base_tx_per_second = (WEIGHT_REF_TIME_PER_SECOND as u128) / base_weight;
+		base_tx_per_second * base_tx_fee()
+	}
 }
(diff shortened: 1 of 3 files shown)


stiiifff:
@hbulgarini Re. the `FixedRateOfFungible` for `Roc`, does this mean we need to allow teleports ? Or create a derived asset and add it to the Asset Registry ?

valentinfernandez1 (approved):

hbulgarini:
> @hbulgarini Re. the `FixedRateOfFungible` for `Roc`, does this mean we need to allow teleports ? Or create a derived asset and add it to the Asset Registry ?

Create a derived asset and add it to the asset registry

PR #222, opened 2023-06-09
Add initial setup of the XCM emulator

hbulgarini:
This PR adds a first version of the XCM emulator into Trappist.

Still this can be considered as a WIP since there are some warnings for unused variables and functions that will be used later when we migrate the use cases from the XCM Simulator into this new testing tool.

@evilrobot-01 : you review is not strictly necessary for this PR but in case you can point something out to improve it's more than welcome.

This PR contains work from the PR #221 , so we might want to merge that one before starting to review this one.


evilrobot-01:
FYI - this is availabe in cumulus now (as of last month), so could just add as a dependency rather than including all the source in this repo.

https://github.com/paritytech/cumulus/tree/master/xcm/xcm-emulator

hbulgarini:
This should be discarded since the replacement of the new XCM emulator from the cumulus repo.

Issue #212, opened 2023-06-06
Remove restriction from asset registry.


hbulgarini:
At the moment only assets from a pallet instance can be reserved transferred from any chain to trappist:

```rust
			ensure!(
				matches!(
					asset_multi_location,
					MultiLocation {
						parents: 1,
						interior: Junctions::X3(Parachain(_), PalletInstance(_), GeneralIndex(_))
					}
				),
				Error::<T>::WrongMultiLocation
			);
```

This check is too restrictive and for example it does not let to register the case of receiving a native token from any chain and minting it using the asset pallet on trappist.

valentinfernandez1:
We could make the restriction more flexible to ensure that the asset comes either from a parachain directly meaning is the native token or a general index meaning is an id for an asset.

This could be done by ensuring that the last item of the junctions is either of those options. Here's an example of this that could work.

```rust
// verify MultiLocation is valid
ensure!(Self::valid_location(&asset_multi_location), Error::<T>::WrongMultiLocation);
```
```rust
impl<T: Config> Pallet<T> {
	fn valid_location(location: &MultiLocation) -> bool {
		let last_junction = location.last();

		if let Some(junction) = last_junction {
			match junction {
				xcm::v1::Junction::Parachain(_) | xcm::v1::Junction::GeneralIndex(_) => true,
				_ => false,
				}
		} else {
			false
		}
	}
}
```

stiiifff:
@valentinfernandez1 I would prefer a more exhaustive check on the location .. let's not take shortuct (I know `register_reserver_asset` must be called by a privileged origin but still ..).

valentinfernandez1:
That's the challenge here, adding a restriction to the format of the multilocation that is not to restrictive like the one we currently have, but that it's also exhaustive enough to block requests with an unsupported multilocation is not as simple.

As I see it we have different alternatives:
+ Remove the restriction all together as suggested by hector
+ Have a generalized restriction that can block most incorrectly formatted requests, like the one I previously listed.
+ Have a set list of accepted multilocation formats that are allowed and compare the received multilocation to verify that matches with that format.

That's why I suggested [this](https://github.com/paritytech/trappist/issues/212#issuecomment-1582991132) implementation since it feels like a middle ground and we can tweak it to be as flexible or as restrictive as we want.

valentinfernandez1:
[...]

PR #224, opened 2023-06-09
Validate `MultiLocation` for asset-registry 

valentinfernandez1:
Improves the validation for the Multilocation received in `register_reserve_asset` inside the Asset-Registry pallet as discussed in #212.

### DIFF
--- a/pallets/asset-registry/src/lib.rs
+++ b/pallets/asset-registry/src/lib.rs
@@ -39,8 +39,8 @@ pub mod pallet {
 	use frame_system::pallet_prelude::*;
 
 	use xcm::latest::{
-		Junction::{GeneralIndex, PalletInstance, Parachain},
-		Junctions, MultiLocation,
+		Junction::{AccountId32, AccountKey20, GeneralIndex, PalletInstance},
+		MultiLocation,
 	};
 
 	#[pallet::pallet]
@@ -113,16 +113,7 @@ pub mod pallet {
 			);
 
 			// verify MultiLocation is valid
-			ensure!(
-				matches!(
-					asset_multi_location,
-					MultiLocation {
-						parents: 1,
-						interior: Junctions::X3(Parachain(_), PalletInstance(_), GeneralIndex(_))
-					}
-				),
-				Error::<T>::WrongMultiLocation
-			);
+			ensure!(Self::valid_location(&asset_multi_location), Error::<T>::WrongMultiLocation);
 
 			// register asset_id => asset_multi_location
 			AssetIdMultiLocation::<T>::insert(asset_id, &asset_multi_location);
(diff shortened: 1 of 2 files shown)


pallets/asset-registry/src/lib.rs
```diff
@@ -153,6 +147,35 @@ pub mod pallet {
...
 	}
 
+	impl<T: Config> Pallet<T> {
+		//Validates that the location points to an asset (Native, Frame based, Erc20) as described
+		// in the xcm-format:  https://github.com/paritytech/xcm-format#concrete-identifiers
+		fn valid_asset_location(location: &MultiLocation) -> bool {
+			let (split_multilocation, last_junction) = location.clone().split_last_interior();
+
+			if let Some(junction) = last_junction {
+				match junction {
```
hbulgarini, line 157: Alex made this refactor using the `maches!` macro which i think makes the code much better in terms of quality:

https://github.com/paritytech/trappist/pull/223/files#diff-d7c7fdfd2cdd4c5fc7b3d6df2550952503740f1819f11331adc24591a8b68e65R81

pallets/asset-registry/src/tests.rs
```diff
@@ -29,84 +29,189 @@ const STATEMINE_ASSET_MULTI_LOCATION: MultiLocation = MultiLocation {
...
+					LOCAL_ASSET_ID,
+					STATEMINE_ASSET_MULTI_LOCATION,
+				),
+				Error::<Test>::AssetAlreadyRegistered
+			);
+		});
+	}
+
+	#[test]
+	fn valid_locations_succced() {
```
hbulgarini, line 92: can we add one more tests to validate relay chain native token?

`MultiLocation { parents: 1, interior: Here `

hbulgarini (changes requested):

pallets/asset-registry/src/tests.rs
valentinfernandez1: Using `Here` will not work since the check is based on how this is suggested by the XCM-Format.

> chain/PalletInstance(<id>) for a Frame chain with a single-asset pallet instance (such as an instance of the Balances pallet).

In order for that to work it would need to be `MultiLocation { parents: 1, interior: PalletInstance(4)}` where the id 4 references the rococo pallet balances.

So I'm not sure whether we should tweak the validation or maybe I`m missing something

pallets/asset-registry/src/lib.rs
valentinfernandez1: Thanks, this does improve the code and reduces a lot of nesting.

pallets/asset-registry/src/tests.rs
hbulgarini: But in PolkadotJS we are allowed to choose `Here` for the Concrete identifiers.

Screenshot 2023-06-10 at 08 24 25

My guess is that it should work

valentinfernandez1: It's fixed with the commit I added yesterday.
```rust
        let (split_multilocation, last_junction) = location.clone().split_last_interior();

	let check = matches!(
		last_junction,
		Some(AccountId32 { .. }) |
			Some(AccountKey20 { .. }) |
			Some(PalletInstance(_)) |
			None
	);
```
If the Junctions is `Here` the function `split_interior()` will return `None`.

hbulgarini (approved):
looks great!

Issue #197, opened 2023-05-25
Add `try_runtime` to Trappist


hbulgarini:
```[tasklist]
### TODO
- [x] Integrate try-runtime into trappist runtime
- [x] Integrate try-runtime into the node cli
- [ ] Create try-runtime CI workflow for releases
- [x] We will need a node to scrape live state from  an RPC node i.e. [wss://rococo-trappist-rpc.polkadot.io:443](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frococo-trappist-rpc.polkadot.io#/explorer)
- [x] Try-runtime bot
- [ ] Documentation
```

The Polkadot CI for try-runtime is located here:
- https://github.com/paritytech/polkadot/blob/master/scripts/ci/gitlab/pipeline/check.yml#L72-L76

Looks like try-runtime-bot is installed: https://github.com/paritytech/security/issues/12

brunopgalvao:
A try-runtime RPC node has been configured for us: 
- wss://rococo-trappist-try-runtime-node.parity-chains.parity.io:443
- https://cloudflare-ipfs.com/ipns/dotapps.io/?rpc=wss%3A%2F%2Frococo-trappist-try-runtime-node.parity-chains.parity.io#/explorer

Tested with this [PR](https://github.com/paritytech/trappist/pull/239) and it works!
```
./node-try-runtime \
    try-runtime \
    --runtime runtime-try-runtime.wasm \
    -lruntime=debug \
    on-runtime-upgrade \
    live --uri wss://rococo-trappist-try-runtime-node.parity-chains.parity.io:443
2023-06-23 12:16:04.421  INFO main remote-ext: since no at is provided, setting it to latest finalized head, 0xc2136fbc8067ffa76b4da8a888cd1bfbc44d3dc0b468bac5caf4a9895910c3e4    
2023-06-23 12:16:04.424  INFO main remote-ext: since no prefix is filtered, the data for all pallets will be downloaded    
2023-06-23 12:16:04.582  INFO main remote-ext: scraping key-pairs from remote at block height 0xc2136fbc8067ffa76b4da8a888cd1bfbc44d3dc0b468bac5caf4a9895910c3e4    
2023-06-23 12:16:05.766  INFO main remote-ext: Querying a total of 4194 keys from prefix , splitting among 10 threads, 420 keys per thread    
2023-06-23 12:16:07.817  INFO main remote-ext: adding data for hashed prefix: , took 3s    
[...]

PR #245, opened 2023-06-29
add try-runtime github workflow

brunopgalvao:
In alignment with https://github.com/paritytech/trappist/issues/197

- [x] ~~Check that the try-runtime feature builds~~
  - We do not need this check on every PR. If anything the try-runtime feature will fail when it runs with the `migration` label
- [x] Run try-runtime on PRs marked with a `migration` label

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -49,6 +49,9 @@ jobs:
       - name: Check Build Trappist node
         run: |
           SKIP_WASM_BUILD=1 cargo check --release --locked
+      - name: Check try-runtime feature
+        run: |
+          cargo check --locked --features try-runtime --all
 
   test:
     needs: lint
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -49,9 +49,6 @@ jobs:
       - name: Check Build Trappist node
         run: |
           SKIP_WASM_BUILD=1 cargo check --release --locked
-      - name: Check try-runtime feature
-        run: |
-          cargo check --locked --features try-runtime --all
 
   test:
     needs: lint
--- a/.github/workflows/try-runtime.yml
+++ b/.github/workflows/try-runtime.yml
@@ -2,7 +2,6 @@ name: Run try-runtime on storage migration label
 
 on:
   pull_request:
-  issues:
     types:
       - labeled
 
--- a/.github/workflows/try-runtime.yml
+++ b/.github/workflows/try-runtime.yml
@@ -29,7 +29,7 @@ jobs:
           key: ${{ runner.os }}-cargo-release-${{ env.POLKA_VERSION }}-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: ${{ runner.os }}-cargo-release-${{ env.POLKA_VERSION }}
 
-      - run: 
+      - run: |
           echo "Found label runtime_migration. Running tests"
           export RUST_LOG=remote-ext=debug,runtime=debug
           echo "---------- Running try-runtime for Trappist ----------"
--- a/.github/workflows/try-runtime.yml
+++ b/.github/workflows/try-runtime.yml
@@ -1,4 +1,4 @@
-name: Run try-runtime on storage migration label
+name: Test storage migration
 
 on:
   push:
@@ -9,7 +9,7 @@ on:
       - labeled
 
 jobs:
-  run_if_label_matches:
+  try_runtime:
     if: github.event.label.name == 'migration'
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/try-runtime.yml
+++ b/.github/workflows/try-runtime.yml
@@ -38,7 +38,6 @@ jobs:
           echo "Found label runtime_migration. Running tests"
           export RUST_LOG=remote-ext=debug,runtime=debug
           echo "---------- Running try-runtime for Trappist ----------"
-          cargo build --locked --release -p trappist-runtime
           cargo run --locked --release --features try-runtime try-runtime \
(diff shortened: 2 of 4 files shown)


brunopgalvao:
We may want to consider doing this for every commit as this is what this PR describes:
- https://github.com/paritytech/polkadot/pull/7422

.github/workflows/try-runtime.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+
+      - name: Setup worker
+        uses: "./.github/templates/setup-worker"
+
+      - name: Cache Build artefacts
+        uses: actions/cache@v3
```
kalaninja, line 20: While you can benefit from using cache here, you are doing a very specific build here. So I suggest that you do a cache [restore](https://github.com/actions/cache/blob/main/restore/README.md) here without saving the resulting artifacts

.github/workflows/try-runtime.yml
```diff
@@ -0,0 +1,40 @@
...
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-release-${{ env.POLKA_VERSION }}-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-cargo-release-${{ env.POLKA_VERSION }}
+
+      - run: 
+          echo "Found label runtime_migration. Running tests"
+          export RUST_LOG=remote-ext=debug,runtime=debug
+          echo "---------- Running try-runtime for Trappist ----------"
+          time cargo build --locked --release -p trappist-runtime
```
kalaninja, line 35: I don't see why you need to build this runtime explicitly since the very next line will do it anyway. If you want to compile only trappist runtime, you can set it up [this](https://github.com/paritytech/trappist/blob/6168897c8811e7ea2f637b2f041e7313cb05f8e2/.github/workflows/check.yml#L80) way

.github/workflows/try-runtime.yml
```diff
@@ -0,0 +1,40 @@
...
+          time cargo run --locked --release --features try-runtime try-runtime \
```
kalaninja, line 36: I am not sure if we need to use `time` command here. job logs already have timestamps.

brunopgalvao: Agreed. Removed.

.github/workflows/try-runtime.yml
brunopgalvao: For this purpose we do not need the `stout-runtime`.

I tried:
```
cargo build --workspace --exclude stout-runtime  --locked --release
```
And for some reason it still builds the `stout-runtime`:
```
   Compiling rococo-runtime v0.9.37 (https://github.com/paritytech/polkadot?branch=release-v0.9.37#64572398)
   Compiling polkadot-runtime v0.9.37 (https://github.com/paritytech/polkadot?branch=release-v0.9.37#64572398)
   Compiling kusama-runtime v0.9.37 (https://github.com/paritytech/polkadot?branch=release-v0.9.37#64572398)
   Compiling trappist-runtime v1.0.0 (/home/bruno/src/trappist/runtime/trappist)
   Compiling stout-runtime v1.0.0 (/home/bruno/src/trappist/runtime/stout)
  Building [====================>  ] 1041/1106: stout-runtime(build), rococo-runtime(build), polkadot-runtime(build), trapp...
```
Any thoughts?

brunopgalvao: I believe you need to compile the runtime explicitly because on the next line when we run `cargo run ...` we have to pass in the runtime wasm:
```
--runtime ./target/release/wbuild/trappist-runtime/target/wasm32-unknown-unknown/release/trappist_runtime.wasm
```

.github/workflows/check.yml
```diff
@@ -49,6 +49,9 @@ jobs:
       - name: Check Build Trappist node
         run: |
           SKIP_WASM_BUILD=1 cargo check --release --locked
+      - name: Check try-runtime feature
+        run: |
+          cargo check --locked --features try-runtime --all
```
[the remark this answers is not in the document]
brunopgalvao: I agree. Removing it for now.

.github/workflows/try-runtime.yml
brunopgalvao: You are right. I am not sure why Polkadot has it this way:
- https://github.com/paritytech/polkadot/blob/6a71a7ef7dbe4968db09d94305ae28d9bd95782d/scripts/ci/gitlab/pipeline/check.yml#L71

If we are calling `cargo run..` on the very next line that should be sufficient.

brunopgalvao:
We also want this workflow to run on subsequent pushes.

.github/workflows/try-runtime.yml
ggwpez: Yea i dont think we need the first line. Can you remove it and try?

brunopgalvao: hmm looks like it worked

kalaninja: > If we are calling `cargo run..` on the very next line that should be sufficient.

Like I said in the first comment.

> I tried:
>
> cargo build --workspace --exclude stout-runtime  --locked --release
>
> And for some reason it still builds the stout-runtime:

You only excluded the stout runtime from the workspace. You need to set feature flags properly.

kalaninja: try this: 
`cargo run -p trappist --locked --release --no-default-features --features trappist/trappist-runtime,try-runtime try-runtime --runtime ...`

kalaninja (approved):

stiiifff (approved):

Issue #231, opened 2023-06-13
[0.9.42] Remove workaround in command-bot


kalaninja:
Because nightly version was updated on `ci-linux` substrate-based chains face `UnknownOpcode(192)` error in wasm runtimes. A [workaround](https://github.com/paritytech/command-bot-scripts/blob/d39e2afd024f5053696954d19e882d1d5d619be5/commands/bench/lib/bench-all-trappist.sh#L7) was introduced. This workaround can be removed after Trappist is migrated to 0.9.42+, where this issue was fixed.

PR #250, opened 2023-07-20
Release 0.9.42 to main.

hbulgarini:
- upgrade polkadot release codebase to 0.9.42
- upgrade rust toolchain to nightly-2023-05-22
- upgrade weights to use `from_parts` constructor instead of temporary `from_ref_time`

next steps after merge:
- #231 
- #251


templates/xcm-bench-template.hbs
```diff
@@ -19,33 +23,45 @@ use sp_std::marker::PhantomData;
...
-			{{#if (ne benchmark.base_writes "0")}}
-			.saturating_add(T::DbWeight::get().writes({{benchmark.base_writes}} as u64))
-			{{/if}}
-			{{#each benchmark.component_writes as |cw|}}
-			.saturating_add(T::DbWeight::get().writes(({{cw.slope}} as u64).saturating_mul({{cw.name}} as u64)))
-			{{/each}}
-	}
-	{{/each}}
-}
+		impl<T: frame_system::Config> WeightInfo<T> {
```
kalaninja, line 26: something weird with indentations here

kalaninja:
I was also curious why you changed `MaxCodeLen` and `CallStack` in `pallet_contracts::Config` and probably is worth mentioning  [this](https://github.com/paritytech/substrate/pull/12993) pr containing information on where these numbers come from, what they affect and how they can be adjusted.

node/src/service.rs
```diff
@@ -280,32 +278,26 @@ where
...
 	)
 	.await
-	.map_err(|e| match e {
-		RelayChainError::ServiceError(polkadot_service::Error::Sub(x)) => x,
-		s => s.to_string().into(),
-	})?;
-
-	let block_announce_validator =
-		BlockAnnounceValidator::new(relay_chain_interface.clone(), para_id);
+	.map_err(|e| sc_service::Error::Application(Box::new(e) as Box<_>))?;
```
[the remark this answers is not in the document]
valentinfernandez1: This was part of #171 and was mostly based on the refactor to the [parachain template v0.9.40 update](https://github.com/substrate-developer-hub/substrate-parachain-template/pull/153)

runtime/trappist/src/constants.rs
```diff
@@ -59,8 +62,38 @@ pub mod fee {
...
+		fn weight_to_fee(weight: &Weight) -> Self::Balance {
+			let ref_time = Balance::saturated_from(weight.ref_time());
+			let proof_size = Balance::saturated_from(weight.proof_size());
+
+			let ref_polynomial = RefTimeToFee::polynomial();
+			let proof_polynomial = ProofSizeToFee::polynomial();
+
+			// Get fee amount from ref_time based on the RefTime polynomial
+			let ref_fee: Balance =
+				ref_polynomial.iter().fold(0, |acc, term| term.saturating_eval(acc, ref_time));
```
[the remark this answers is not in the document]
valentinfernandez1: Agreed, this was a temporary solution but now this should be changed to use FeePolynomial similar to [asset-hub](https://github.com/paritytech/cumulus/blob/1d652af24a6f9fb1d97751b4560529bb8c83217d/parachains/runtimes/assets/asset-hub-polkadot/src/constants.rs#L58)

valentinfernandez1:
> I was also curious why you changed `MaxCodeLen` and `CallStack` in `pallet_contracts::Config` and probably is worth mentioning [this](https://github.com/paritytech/substrate/pull/12993) pr containing information on where these numbers come from, what they affect and how they can be adjusted.

This also comes from the update to v0.9.40, and it follows https://github.com/paritytech/cumulus/pull/2108

runtime/trappist/src/lib.rs
```diff
@@ -213,17 +214,21 @@ impl pallet_authorship::Config for Runtime {
...
 	type ExistentialDeposit = ConstU128<EXISTENTIAL_DEPOSIT>;
 	type AccountStore = System;
-	type WeightInfo = pallet_balances::weights::SubstrateWeight<Runtime>;
+	type WeightInfo = weights::pallet_balances::WeightInfo<Runtime>;
+	type MaxLocks = ConstU32<50>;
 	type MaxReserves = ConstU32<50>;
 	type ReserveIdentifier = [u8; 8];
+	type HoldIdentifier = ();
+	type FreezeIdentifier = ();
+	type MaxHolds = ConstU32<0>;
```
valentinfernandez1, line 230: Is it okey that we keep this values at `ConstU32<0>`. Not completly sure what would be the best alternative

.github/templates/free-space/action.yaml
```diff
@@ -0,0 +1,23 @@
+name: free-space
```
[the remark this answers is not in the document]
kalaninja: I think this just came from main branch. I added this to free space on ubuntu runners, because the previous one that we were using wasn't good enough.

runtime/trappist/src/lib.rs
kalaninja: I think currently `construct_runtime` macro generates `RuntimeFreezeReason` and `RuntimeHoldReason` enums based on `#[pallet::composite_enum]` attributes in pallets. These should be used as ***Identifiers here. 
As freezes are the new locks and reserves were deprecated in favor of holds I think the corresponding limits can be reused. 
See:
- https://github.com/paritytech/substrate/pull/12951
- https://github.com/paritytech/substrate/blob/master/frame/support/procedural/src/construct_runtime/expand/hold_reason.rs
- https://github.com/paritytech/substrate/blob/master/frame/support/procedural/src/construct_runtime/expand/freeze_reason.rs

But anyway this is not critical for the migration and can be addressed in a separate pr.

node/src/service.rs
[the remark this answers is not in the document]
kalaninja: What is the point in casting Box to Box? this is not needed in the parachain template either. It can be needed sometimes, but this doesn't seem to be the case here.

valentinfernandez1:
Just added the fix to the WeightsToFee. The commit also includes some formatting fixes.

hbulgarini:
@kalaninja we are ready to merge right ?

runtime/trappist/src/xcm_config.rs
```diff
@@ -97,26 +106,21 @@ pub type LocalAssetTransactor = CurrencyAdapter<
 	// Our chain's account ID type (we can't get away without mentioning it explicitly):
 	AccountId,
 	// We don't track any teleports of `Balances`.
-	(),
+	CheckAccount,
```
stiiifff, line 109: Is this to support the use case with the AH ?

Cargo.toml
```diff
@@ -44,147 +44,149 @@ stout-runtime = { path = "runtime/stout" }
 jsonrpsee = { version = "0.16.2" }
```
stiiifff, line 44: Need to update `clap`, `serde`, `hex-literal`, `scale-info` versions too.

stiiifff (changes requested):

runtime/trappist/src/xcm_config.rs
hbulgarini: No, however I would do it in a different PR

Cargo.toml
kalaninja: done

runtime/trappist/src/xcm_config.rs
kalaninja: I think it initially came from [here](https://github.com/paritytech/trappist/commit/6cd2ee699dcf7851a8dbb96eccf217f8459e75ab#diff-762c8a6849a94f93f63b9751aba00bbdb14ad9f5794eb7d14c964271bf1fbe68R101)

hbulgarini: > Is this to support the use case with the AH ?

Correct, this needed to keep track of the in/out coming teleports for Trappist.

 To be honest when we add this one we were not very mindful why we were doing it , but now with the AH use case it's 100% clear and needed.

stiiifff: @hbulgarini Let's at least update the comment then, because it is not inline anymore with the change.

hbulgarini: > @hbulgarini Let's at least update the comment then, because it is not inline anymore with the change.

Actually, i think the configuration was wrong. The only place where the `CheckAccount` was needed is on the CurrencyTransactor that is the one that takes care for the HOP teleport to AH (teleport of native token).
Then for FungibleTransactors (meaning assets) this is not required since for the moment we don't plan to teleport any asset to any location.

stiiifff (approved):

stiiifff:
LGTM 🚀️

PR #253, opened 2023-08-01
Bump spec_version to 9420 and transaction_version to 2

hbulgarini:


runtime/trappist/src/lib.rs
```diff
@@ -134,7 +134,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("trappist-rococo"),
 	impl_name: create_runtime_str!("trappist-rococo"),
 	authoring_version: 1,
-	spec_version: 9370,
+	spec_version: 9420,
```
stiiifff, line 137: Shouldn't we bump `transaction_version` as well ?

hbulgarini: done.

stiiifff (approved):

PR #254, opened 2023-08-02
Remove lockdown mode from GenesisConfig for local

hbulgarini:
There is no point on starting Trappist locally with the lockdown mode activated. Since it was added many users thinks the chain does not work when it's launched locally. 

For this reason, i have updated the local chain spec of Trappist to start the chain without the lockdown mode activated.


metricaez:
Tested locally and working. I was one of the users that took a while to figure out it was locked so can relate.

hbulgarini:
> Tested locally and working. I was one of the users that took a while to figure out it was locked so can relate.

amazing! could you please approve :)

metricaez (approved):

PR #255, opened 2023-08-02
Removing CheckAccount from currency adapter

hbulgarini:


metricaez (approved):
Change needed at the time being for incoming teleports.

PR #257, opened 2023-08-02
Changing XCM config for allow teleport from and to AH

metricaez:
Basic changes that need to be implemented for allowing teleporting from and to AH from Trappist. Remove filter, add AH as trusted teleporter and fix Currency adapter.

### DIFF
--- a/runtime/trappist/src/xcm_config.rs
+++ b/runtime/trappist/src/xcm_config.rs
@@ -331,7 +331,7 @@ impl pallet_xcm::Config for Runtime {
 	type ExecuteXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
 	type XcmExecuteFilter = Everything;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type XcmTeleportFilter = Nothing;
+	type XcmTeleportFilter = Everything;
 	type XcmReserveTransferFilter = Everything;
 	type Weigher = WeightInfoBounds<
 		crate::weights::xcm::TrappistXcmWeight<RuntimeCall>,
--- a/runtime/trappist/src/xcm_config.rs
+++ b/runtime/trappist/src/xcm_config.rs
@@ -203,6 +203,7 @@ match_types! {
 	};
 }
 
+
 pub type Barrier = DenyThenTry<
 	DenyReserveTransferToRelayChain,
 	(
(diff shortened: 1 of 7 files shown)


runtime/trappist/src/xcm_config.rs
```diff
@@ -331,7 +337,7 @@ impl pallet_xcm::Config for Runtime {
 	type ExecuteXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
 	type XcmExecuteFilter = Everything;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type XcmTeleportFilter = Nothing;
+	type XcmTeleportFilter = Everything;
```
hbulgarini, line 340: hmmm i think we should provide `TrustedTeleporters` here instead of everything right?

hbulgarini (changes requested):

runtime/trappist/src/xcm_config.rs
metricaez: Since this is mostly focused on what could or should try to be teleported out of Trappist I didn't apply any criteria. Should we limit this ?

metricaez: Makes all the sense, will implement 🤝

hbulgarini:
@metricaez: is looking very good! however there is a test failing in the asset registry, could you please take a look?

runtime/trappist/src/xcm_config.rs
metricaez: Implemented allowing only Native assets to be teleported out of Trappist.

metricaez:
Changes implemented:

- Allow teleports from Trappist, implemented a filter that only allows teleporting HOP.
- Allow teleports of HOP from AH to Trappist
- Changed Stout and Trappist Reserves to check that origin is AH but not the asset itself as it generated conflicts with the multilocation of foreign assets. Asset check is done by Asset Registry.
- Added ParachainIDs as valid junction for asset registration on Asset Registry since it's needed for foreign assets.
- Implement HOP Trader on Stout to buy execution with arbitrary value.

@hbulgarini please give it a look and let me know any change you think is needed, otherwise good to merge from my side.

runtime/trappist/src/xcm_config.rs
```diff
@@ -245,13 +252,24 @@ pub struct ReserveAssetsFrom<T>(PhantomData<T>);
...
-			}
+		log::trace!(target: "xcm::AssetsFrom", "prefix: {:?}, origin: {:?}, asset: {:?}", prefix, origin, asset);
+		&prefix == origin
+	}
+}
+
+pub struct OnlyTeleportNative;
+impl Contains<(MultiLocation, Vec<MultiAsset>)> for OnlyTeleportNative {
+	fn contains(t: &(MultiLocation, Vec<MultiAsset>)) -> bool {
+		if t.1.len() != 1 {
```
hbulgarini, line 263:
```suggestion
t.1.iter().any(|asset| {
    log::trace!(target: "xcm::OnlyTeleportNative", "Asset to be teleported: {:?}", asset);
    match asset {
        MultiAsset { id: xcm::latest::AssetId::Concrete(asset_loc), fun: Fungible(_a) } =>
            matches_prefix(&SelfReserve::get(), asset_loc),
        _ => false,
    }
})
```

I'm in personal conflict with this line: Your proposal is much better in terms of performance and cost of execution since it does not iterate. On the other side, i don't like how the `if t.1.len() != 1` reads and also i have a feeling that this will not escalate much if we need to add another scenario to the teleport ( not clear how a new scenario would fit here).

@stiiifff  what do you think?

hbulgarini (approved):

PR #275, opened 2023-09-06
Withdraw derivative on origin

metricaez:
Implementation of `do_withdraw_and_teleport` call from #266 that also withdraws a derivative asset on origin chain 

For the sake of simplicity on wording:

Origin chain: Trappist
Destiny chain: Asset Hub

Pros: 

Users cannot longer exploit Trappist’s SovereignAccount on Asset Hub as they would need to have the same amount of funds as derivative on their Trappist's account than they would want to use for buying execution on Asset Hub.

ROCs used for buying execution on Asset Hub are not longer trapped and they are refunded as ROC on Asset Hub, not as the derivatives on Trappist.

Cons: 

Users must have derivative of ROC on Trappist apart from the HOP that they want to teleport. This could be done through a simple `ReserveAssetTransfer`

TODO: 
Refund surplus as derivative on origin chain to origin account. 

Known issue:
The amount of native asset that is minted through `ReceiveTeleportedAsset` does not take in consideration the amount used for buying execution on origin chain, therefore it is "free" from the origin chain native asset perspective.

### DIFF
--- a/pallets/withdraw-teleport/src/lib.rs
+++ b/pallets/withdraw-teleport/src/lib.rs
@@ -202,12 +202,14 @@ impl<T: Config> Pallet<T> {
 			// There are currently no limitations on the amount of fee assets to withdraw.
 			// Since funds are withdrawn from the Sovereign Account of the origin, chains must be
 			// aware of this and implement a mechanism to prevent draining.
-			WithdrawAsset(fee_asset),
+			WithdrawAsset(fee_asset.clone()),
 			BuyExecution { fees, weight_limit },
 			ReceiveTeleportedAsset(foreing_assets.clone()),
 			// Intentionally trap ROC to avoid exploit of draining Sovereing Account
 			// by depositing withdrawn ROC on beneficiary.
 			DepositAsset { assets: MultiAssetFilter::Definite(foreing_assets), beneficiary },
+			RefundSurplus,
+			DepositAsset { assets: MultiAssetFilter::Definite(fee_asset), beneficiary },
 		]);
 
 		// Temporarly hardcode weight.
--- a/pallets/withdraw-teleport/src/lib.rs
+++ b/pallets/withdraw-teleport/src/lib.rs
@@ -24,7 +24,7 @@
 type BaseXcm<T> = pallet_xcm::Pallet<T>;
 use frame_support::{
 	dispatch::DispatchResult,
-	ensure,
+	ensure, log,
 	traits::{Contains, EnsureOrigin, Get},
 	weights::Weight,
 };
@@ -187,9 +187,11 @@ impl<T: Config> Pallet<T> {
 		//Build the message to execute on origin.
 		let assets: MultiAssets = assets.into();
 		let message: Xcm<<T as frame_system::Config>::RuntimeCall> = Xcm(vec![
+			// Burn the native asset.
 			WithdrawAsset(assets.clone()),
 			BuyExecution { fees: native_asset, weight_limit: Unlimited },
 			BurnAsset(assets),
+			// Burn the fee asset derivative.
 			WithdrawAsset(fee_asset.clone()),
 			BurnAsset(fee_asset.clone()),
 		]);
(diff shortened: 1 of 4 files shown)


pallets/withdraw-teleport/src/lib.rs
```diff
@@ -185,23 +186,29 @@ impl<T: Config> Pallet<T> {
...
 			BuyExecution { fees, weight_limit },
-			ReceiveTeleportedAsset(foreign_assets.clone()),
-			// Intentionally trap ROC to avoid exploit of draining Sovereing Account
-			// by depositing withdrawn ROC on beneficiary.
-			DepositAsset { assets: MultiAssetFilter::Definite(foreign_assets), beneficiary },
+			ReceiveTeleportedAsset(foreing_assets.clone()),
+			// We can deposit funds since they were both withdrawn on origin.
+			DepositAsset { assets: MultiAssetFilter::Definite(foreing_assets), beneficiary },
+			RefundSurplus,
+			DepositAsset { assets: MultiAssetFilter::Definite(fee_asset), beneficiary },
```
joepetrowski, line 211: But you don't know how much of `fee_asset` remains, since part of it was moved into fee payment. I think you need an `AllOf` variant.

metricaez: Ah, I had the misconception that this `MultiAssetFilter` deposited "whatever you have left of x asset" but it takes also in consideration the amount? Will fix.

pallets/withdraw-teleport/src/lib.rs
```diff
@@ -185,23 +186,29 @@ impl<T: Config> Pallet<T> {
 
 		//Build the message to execute on origin.
 		let assets: MultiAssets = assets.into();
-		let message: Xcm<<T as frame_system::Config>::RuntimeCall> =
-			Xcm(vec![WithdrawAsset(assets.clone()), BurnAsset(assets)]);
+		let message: Xcm<<T as frame_system::Config>::RuntimeCall> = Xcm(vec![
+			// Burn the native asset.
+			WithdrawAsset(assets.clone()),
+			BuyExecution { fees: native_asset, weight_limit: Unlimited },
+			BurnAsset(assets),
```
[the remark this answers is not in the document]
metricaez: Indeed, and this also applies for the amount to be minted on AH on the sent `ReceiveTeleportedAsset` as stated on the known issue [here](https://github.com/paritytech/trappist/pull/275#issue-1884789887). 

Will dig into this.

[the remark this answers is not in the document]
metricaez: Update: Fixed with `SetFeesMode`, fees are paid from origin on-chain account funds so amount withdrawn is amount to be burned and minted on destination.

stiiifff:
@joepetrowski Could you take another (hopefully final) look ? 🙏️

runtime/trappist/src/xcm_config.rs
```diff
@@ -230,7 +230,7 @@ parameter_types! {
 		0u128
 	);
 	/// Roc = 7 RUSD
-	pub RocPerSecond: (xcm::v3::AssetId, u128,u128) = (MultiLocation::parent().into(), default_fee_per_second() * 70, 0u128);
+	pub RocPerSecond: (xcm::v3::AssetId, u128,u128) = (MultiLocation::new(1,Here).into(), default_fee_per_second() * 70, 0u128);
```
joepetrowski, line 233: Shouldn't `fmt` catch things like this?
```suggestion
	pub RocPerSecond: (xcm::v3::AssetId, u128, u128) = (MultiLocation::new(1, Here).into(), default_fee_per_second() * 70, 0u128);
```

joepetrowski (approved):

runtime/trappist/src/xcm_config.rs
Moliholy: Recently a few formatting rules were changed since they were only available in nightly versions of Rust. That might be the reason.

PR #294, opened 2023-09-19
ci(worker): use stable rust

evilrobot-01:
Updates setup worker action to use stable rust, which is now the default as per the toolchain file.


evilrobot-01:
Closing in favour of the other PR

PR #296, opened 2023-09-28
Withdraw-teleport-benchmark

metricaez:


stiiifff (approved):

pallets/withdraw-teleport/src/lib.rs
```diff
@@ -101,7 +102,28 @@ pub mod pallet {
...
+				Ok(assets) => {
+					use sp_std::vec;
+					let mut message = Xcm(vec![
+						WithdrawAsset(native_assets.clone()),
+						SetFeesMode { jit_withdraw: true },
+						BurnAsset(native_assets),
+						WithdrawAsset(assets.clone()),
+						BurnAsset(assets),
+					]);
+					T::Weigher::weight(&mut message).map_or(Weight::MAX, |w| <T as pallet::Config>::WeightInfo::withdraw_and_teleport().saturating_add(w))
```
hbulgarini, line 122: I'm not 100% sure about this but here is my concern:

You are computing the weight for the XCM messages and once you have the result you sum that weight to the benchmarking for the extrinsic weight.  Aren't these XCM messages charged already on the XCM Executor when the messages are processed? In other terms , are you not charging twice the XCM messages?

pallets/withdraw-teleport/src/weights.rs
```diff
@@ -0,0 +1,85 @@
...
+	/// Storage: `ParachainSystem::HostConfiguration` (r:1 w:0)
+	/// Proof: `ParachainSystem::HostConfiguration` (`max_values`: Some(1), `max_size`: None, mode: `Measured`)
+	/// Storage: `ParachainSystem::PendingUpwardMessages` (r:1 w:1)
+	/// Proof: `ParachainSystem::PendingUpwardMessages` (`max_values`: Some(1), `max_size`: None, mode: `Measured`)
+	fn withdraw_and_teleport() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `177`
+		//  Estimated: `3642`
+		// Minimum execution time: 87_000_000 picoseconds.
+		Weight::from_parts(88_000_000, 3642)
```
hbulgarini, line 56: for some reason the ref_time is different from the benchmark result:

https://github.com/paritytech/trappist/pull/296/files#diff-220649b2a41e8072e94f45ddbe66886b3fbea765f3546d9aef12367a0681461cR50

Even if this will not be final results, because they have to be properly benchmarked with the bot i'm curious about this.

pallets/withdraw-teleport/src/lib.rs
metricaez: I stuck to the pattern used on [pallet-xcm](https://github.com/paritytech/polkadot/blob/52209dcfe546ff39cc031b92d64e787e7e8264d4/xcm/pallet-xcm/src/lib.rs#L806) and traced relevant [PR](https://github.com/paritytech/polkadot/issues/7182) even though I agree it's not fully clear to me either. Already discussed this with xcm team and agreed on keeping it until they address it further just to be sure on not being undercharging. 
What I do also realize is that if I want to fully stick to the pattern I should also add `pallet-xcm` `send` weight as it is also called inside the extrinsic, this is opaque on the case of `teleport` and `reserve` as it is done inside the instructions.

runtime/stout/src/xcm_config.rs
```diff
@@ -198,30 +197,19 @@ match_types! {
...
-					AllowUnpaidExecutionFrom<ParentOrParentsExecutivePlurality>,
-					// Subscriptions for version tracking are OK.
-					AllowSubscriptionsFrom<ParentOrSiblings>,
-				),
-				UniversalLocation,
-				ConstU32<8>,
-			>,
-		),
-	>,
+pub type Barrier = DenyThenTry<
```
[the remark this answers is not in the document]
metricaez: `WithComputedOrigin` is generating issues with `FixedRateOfFungible` `Trader` with `ForeignAssets`. This does not happen to `TrustedAssets` which I have already tested so it doesn't break any previous use case however it does break foreign assets use case and since other teams are using it for testing I would leave it without `WithComputedOrigin` until either this is fixed or we figure out the reason behind it and open a related `issue`

pallets/withdraw-teleport/src/weights.rs
metricaez: Because they are two different runs. 

This one is only used at [mock](https://github.com/paritytech/trappist/blob/e76a63d2ae4bb72628bd0d6674955eee9c55b8e3/pallets/withdraw-teleport/src/mock.rs#L244) Runtime. 

The one that will be used on the runtime itself is the one of the bot benchmark and I had to also run it locally for the [trait](https://github.com/paritytech/trappist/blob/e76a63d2ae4bb72628bd0d6674955eee9c55b8e3/runtime/trappist/src/lib.rs#L654) to be generated not the value itself as it will be overwritten.

templates/frame-weight-template.hbs
```diff
@@ -0,0 +1,121 @@
+{{header}}
```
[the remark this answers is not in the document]
metricaez: This template is only being used for generating the weights that are used on mock runtime, specially due to [this](https://github.com/paritytech/trappist/blob/e76a63d2ae4bb72628bd0d6674955eee9c55b8e3/pallets/withdraw-teleport/src/weights.rs#L38) which is what we are using on our pallets mocks. As far as I understand, the ones used on the actual runtime [do not use it](https://github.com/paritytech/trappist/blob/e76a63d2ae4bb72628bd0d6674955eee9c55b8e3/runtime/trappist/src/weights/pallet_withdraw_teleport.rs) so there should be no issue.

pallets/withdraw-teleport/src/lib.rs
metricaez: What I do also realize is that if I want to fully stick to the pattern I should also add `pallet-xcm` `send` weight as it is also called inside the extrinsic, this is opaque on the case of teleport and reserve as it is done inside the instructions.

hbulgarini (approved):

stiiifff (approved):

PR #304, opened 2023-10-17
Add-xcm-emulator-and-tests

metricaez:
- Implemented emulator from polkadot-sdk
- Migrated tests from simulator to emulator
- Fixed xcm config for reserve transfer from Trappist to Stout use case to pass with current runtimes
- Left duplicated code on tests for making them more explicit as they document the step by step process for each use case
- Two migrated Asset Trap test failing due to current state of runtimes, will open in a new issue as it is not relevant to emulator integration


stiiifff (approved):
Awesome, thanks @metricaez ! LGTM !

PR #305, opened 2023-10-23
feat: add tx-pause and safe-mode pallets

### LINKED ISSUE #279: Replace Lockdown Mode pallet by FRAME SafeMode & TxPause pallets
Trappist runtime includes a custom built Lockdown Mode pallet that provides the ability to temporarily pause transfers & normal operations on a live parachain, which is a valuable mechanism to have at the team’s disposal when an issue / attack occurs.
The Substrate FRAME library now contains the [SafeMode & TxPause pallets](https://github.com/paritytech/substrate/pull/12092) which allow implementation of such a mechanism.

We should prefer the new SafeMode pallet over Trappist’s custom-built one as the former has been more extensively reviewed and audited, and thus, replace the Lockdown Mode pallet by the FRAME SafeMode & TxPause pallets.

The Lockdown Mode pallet should be removed from the Trappist repository, and moved to a separate one, and archived.

- [ ] Integrate the FRAME SafeMode & TxPause pallets with sane defaults in both devnet & mainnet runtimes (disabled by default in both runtimes).
- [ ] SafeMode should only allow balance transfers as whitelisted call. Permissionless operations (enter, extend) should be disabled. Only a privileged origin can enable / disable safe mode: `ForceEnterOrigin`, `ForceExtendOrigin` & `ForceExitOrigin` to be configured as the Root origin (Sudo).
- [ ] TxPause `PauseOrigin` & `UnpauseOrigin` to be configured as the Root origin (Sudo).
- [ ] Remove the Lockdown Mode pallet from the Trappist repository (move it to a separate new repository, and mark it as archived).

Depends on #278

Moliholy:
Closes #279


runtime/trappist/src/constants.rs
```diff
@@ -26,6 +26,7 @@ pub mod currency {
 	pub const CENTS: Balance = constants::currency::CENTS;
 	pub const GRAND: Balance = constants::currency::GRAND;
 	pub const MILLICENTS: Balance = constants::currency::MILLICENTS;
+	pub const DOLLARS: Balance = 100 * CENTS;
```
stiiifff, line 29: To be removed, I don't think this is needed (see comments re. the pallet's config trait's parameters).

runtime/trappist/src/lib.rs
```diff
@@ -654,13 +656,66 @@ impl pallet_withdraw_teleport::Config for Runtime {
...
-	type LockdownDmpHandler = LockdownDmpHandler;
-	type XcmExecutorManager = XcmExecutionManager;
-	type WeightInfo = weights::pallet_lockdown_mode::WeightInfo<Runtime>;
+	type Currency = Balances;
+	type RuntimeHoldReason = RuntimeHoldReason;
+	type WhitelistedCalls = SafeModeWhitelistedCalls;
+	type EnterDuration = EnterDuration;
+	type ExtendDuration = ExtendDuration;
+	type EnterDepositAmount = EnterDepositAmount;
+	type ExtendDepositAmount = ExtendDepositAmount;
```
stiiifff, line 689: The issue mentioned "_Permissionless operations (enter, extend) should be disabled._". IIRC it's possible to disallow the permissionless operations by specifying a zero amount for these params ? Please verify.

runtime/trappist/src/lib.rs
```diff
@@ -654,13 +656,66 @@ impl pallet_withdraw_teleport::Config for Runtime {
...
+			_ => false,
+		}
+	}
+}
+
+parameter_types! {
+	pub const EnterDuration: BlockNumber = 4 * HOURS;
+	pub const EnterDepositAmount: Balance = 2_000_000 * DOLLARS;
+	pub const ExtendDuration: BlockNumber = 2 * HOURS;
+	pub const ExtendDepositAmount: Balance = 1_000_000 * DOLLARS;
```
stiiifff, line 677: To be removed, see below.

runtime/trappist/src/lib.rs
```diff
@@ -700,7 +755,8 @@ construct_runtime!(
 		Uniques: pallet_uniques = 43,
 		Scheduler: pallet_scheduler = 44,
 		Preimage: pallet_preimage = 45,
-		LockdownMode: pallet_lockdown_mode = 46,
+		SafeMode: pallet_safe_mode = 46,
```
stiiifff, line 758: I'm not sure about re-using the same index here. To be on the safe side, I would skip index 46 (and leave a comment). It shouldn't be a problem though because we will increase the runtime `transaction_version` for the next runtime upgrade of Trappist on Rococo. But I will still prefer to err on the safe side on this one.

stiiifff (changes requested):

runtime/trappist/src/lib.rs
Moliholy: You're right, better be on the safe side. Changed in [33b1736](https://github.com/paritytech/trappist/pull/305/commits/33b173617174bd419a5246670ac3699375c25d51).

runtime/trappist/src/lib.rs
Moliholy: Done in [33b1736](https://github.com/paritytech/trappist/pull/305/commits/33b173617174bd419a5246670ac3699375c25d51).

runtime/trappist/src/lib.rs
Moliholy: Very true, I didn't see that. Fixed in [33b1736](https://github.com/paritytech/trappist/pull/305/commits/33b173617174bd419a5246670ac3699375c25d51).

runtime/trappist/src/constants.rs
Moliholy: Done in [33b1736](https://github.com/paritytech/trappist/pull/305/commits/33b173617174bd419a5246670ac3699375c25d51).

stiiifff:
Looks good, thanks @Moliholy !
Let's wait to merge it until #308 has been merged.

stiiifff (approved):

PR #308, opened 2023-10-31
Prepare Trappist v1.1.0 release

stiiifff:


stiiifff:
bot bench-all trappist

stiiifff:
bot bench-all trappist

metricaez:
bot bench-all trappist

stiiifff:
bot bench-all trappist

pgherveou:
You need to set `type MaxHolds` on pallet_balances Config to some non 0 value here, or pallet_contracts fails to hold some balance for uploading new contracts, and you end up with the error StorageDepositNotEnoughFunds from the jobs log. 

Will add some more warn in the pallet_contracts logs to make this more obvious!

metricaez:
bot bench-all trappist

stiiifff:
Ignoring the `try-runtime` check error, as there's effectively is no runtime migration in this release.

Moliholy:
> Ignoring the `try-runtime` check error, as there's effectively is no runtime migration in this release.

I just run the migration and it turns out `pallet-contracts` in at version 12 instead of 15. That's, as you've mentioned, not supposed to be that way, because trappist was migrated in the past to version 15. However, try-runtime says otherwise. I think it'd be wise to further research this issue.

Furthermore, when running try-runtime with the `try-state` flag, I discovered there's data corruption in `pallet-collective`. An extra migration would be needed to solve this issue.

metricaez:
Ignoring try-runtime error as it is non breaking while we debug it.

metricaez (dismissed):

metricaez:
bot bench-all trappist

metricaez (approved):

metricaez (approved):

stiiifff:
LGTM !

PR #312, opened 2023-11-21
Revert "Add-xcm-emulator-and-tests (#304)"

metricaez:
This reverts commit e51c1fead095341210e2a7a5a5971900f476636c.

Decommission emulator es it is generating conflicts in v1.1 with overhead benchmarking and asynchronous backing. Propose to update to v1.1 and forward and test integrations of emulator in more recent versions that might have fixed this issue.  
Tests migrations and implementation is still valid so the process of re integrating it would be much faster.


metricaez:
bot bench-all trappist

metricaez:
@stiiifff I propose we take this path. Even though a new problem appeared.

stiiifff:
Awesome @metricaez, thanks for tracking this down 🙏️ to the latest addition of the XCM emulator. Indeed, let's finalize the v1.1 upgrade, and add the XCM emulator back at a later point.

stiiifff (approved):

PR #315, opened 2023-12-01
feat: upgrade trappist to polkadot-v1.3.0

Moliholy:


Moliholy:
Please note that `pallet-dex` has been removed, as it was causing several problems very difficult to track, even after migrating it to 1.3.0. Instead, we will soon incorporate the [asset conversion pallet](https://github.com/paritytech/trappist/pull/309).

brunopgalvao (dismissed):

runtime/trappist/src/constants.rs
```diff
@@ -16,8 +16,8 @@
 // limitations under the License.
 
 pub mod currency {
-	use kusama_runtime_constants as constants;
 	use polkadot_core_primitives::Balance;
+	use rococo_runtime_constants as constants;
```
stiiifff, line 20: Seemingly innocuous change ... but, if I'm not mistaken, it does change the scale of the currency units 😬️ (`CENTS`, `GRAND`, `MILLICENTS`). Please double-check.

runtime/stout/src/constants.rs
```diff
@@ -16,8 +16,8 @@
 // limitations under the License.
 
 pub mod currency {
-	use kusama_runtime_constants as constants;
 	use polkadot_core_primitives::Balance;
+	use rococo_runtime_constants as constants;
```
stiiifff, line 20: Same remark as for the Trappist runtime.

stiiifff (changes requested):

runtime/trappist/src/constants.rs
Moliholy: `rococo_runtime_constants::currency::UNITS` = 1_000_000_000_000

`rococo_runtime_constants::currency::UNITS` = 1_000_000_000_000

They are the same.

Moliholy: Oh but you are totally right, the other constants are different!

Fixed in [88a0b6a](https://github.com/paritytech/trappist/pull/315/commits/88a0b6a5a8a246bf304a0f815d53f372c71b8323).

Very nice catch!

runtime/stout/src/constants.rs
Moliholy: Fixed in [88a0b6a](https://github.com/paritytech/trappist/pull/315/commits/88a0b6a5a8a246bf304a0f815d53f372c71b8323).

runtime/trappist/src/constants.rs
stiiifff: 🤓️😉️

PR #331, opened 2024-01-21
Asset Hub - Stout NFT transfers

Szegoo:
With cross-chain NFT transfers enabled on asset hub: https://github.com/paritytech/polkadot-sdk/pull/2796 a good demonstration of XCM NFT transfers is enabling it between stout and asset hub.

I will probably look into adding this to Trappist as well, however, that will require some changes to the `asset-registry` pallet so that will be a separate PR.

### DIFF
--- a/runtime/stout/src/lib.rs
+++ b/runtime/stout/src/lib.rs
@@ -467,10 +467,13 @@ parameter_types! {
 	pub const ValueLimit: u32 = 64;
 }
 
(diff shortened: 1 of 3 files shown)


stiiifff (approved):

stiiifff:
@Szegoo Awesome, thanks ! Would love to see it in Trappist so this can be demoed on Rococo.

